PR #123, opened 2012-08-03
Firefox 2 last/indexOf bug detection

dmp42:
In Firefox 2 (Mozilla/5.0 (Macintosh; U; Intel Mac OS X; en-US; rv:1.8.1) Gecko/20061010 Firefox/2.0), this will fail:
[0, 1].indexOf(1, 2) == -1
[0, 1].lastIndexOf(0, -3) == -1

Interestingly, that seem to happen only if the needle is the last element of the array (for indexOf) / the first (for lastIndexOf).

I modified the test slightly to reveal that behavior (for indexOf), and the shim to override indexOf/lastIndexOf in case these fail.


ljharb:
https://github.com/es-shims/Array.prototype.lastIndexOf/blob/019813b31bd451434c971d6cbe403f17eb986045/polyfill.js#L19-L22

PR #300, opened 2015-04-10
User `files` in package.json

SimenB:
This is to avoid installing a whole bunch of unneeded files when installing from `npm`.
Diff:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,10 @@
       "url": "http://github.com/es-shims/es5-shim/raw/master/LICENSE"
     }
   ],
+  "files": [
+    "es5-shim.js",
+    "es5-sham.js"
+  ],
   "main": "es5-shim.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


ljharb:
I'm going to revert this - those files are *needed*.

SimenB:
@ljharb Oh, I dind't know that. for what?

ljharb:
`npm explore es5-shim -- npm install && npm test` should always work for any package.

SimenB:
Hmm, interesting. I've never seen that before.
Do you have a reference to documentation/a discussion that says so?

ljharb:
It's not a requirement, but imo it's conventional since most packages don't worry about what files they contain when they get published to `npm`. I doubt anyone's ever written about it one way or the other tbh.

Disk space is cheap and shouldn't be a primary development concern.

michaelficarra:
@ljharb Just about every project I've worked on uses this field to indicate which files are meant for distribution.

PR #305, opened 2015-05-30
IE8 - Fix #252 'constructor' in Object.defineProperties

### LINKED ISSUE #252: Object.defineProperties(obj, { constructor: .. }} broken in IE8
Properties other than `constructor` seem to work fine. And when using `Object.defineProperty` directly it works fine as well.

I ran into this when using `Object.create` to set up a subclass prototype and restoring the `constructor` property afterwards:

```js
	targetFn.prototype = Object.create( originFn.prototype, {
		// Restore constructor property of targetFn
		constructor: {
			value: targetConstructor,
			enumerable: false,
			writable: true,
			configurable: true
		}
	} );
```

In modern browsers (e.g. latest Chrome) without the es5-shim, `targetFn.prototype.constructor === targetConstructor`. But in IE8 the defineProperties call was a no-op.

```js
// Windows 7 / MSIE 8.0 (browserstack.com)
// Blank html file with es5-shim.js and es5-sham.js (v3.4.0)

var a = {};
a.foo = 1;
Object.defineProperty(a, 'foo', { value: 3 });
a.foo;
//> 2
Object.defineProperty(a, 'constructor', { value: String });
a.constructor + '';
//> function String() { [native code] }

var b = {};
b.foo = 1;
Object.defineProperties(b, { foo: { value: 2 }});
b.foo;
//> 2
Object.defineProperties(b, { constructor: { value: String }});
b.constructor + '';
//> function Object() { [native code] }

// ^ Property was not updated
```

andybry:
Fix #252 

In IE8 the 'constructor' property was not being defined by the Object.defineProperties method of the sham because IE8 does not enumerate the constructor property.

I've tested this change manually against IE8 running on Windows XP, but I wasn't able to work out how you do your automated cross browser testing.

### DIFF
--- a/es5-sham.js
+++ b/es5-sham.js
@@ -386,11 +386,11 @@ if (!Object.defineProperties || definePropertiesFallback) {
             }
         }
 
-        for (var property in properties) {
-            if (owns(properties, property) && property !== '__proto__') {
+        Object.keys(properties).forEach(function (property) {
+            if (property !== '__proto__') {
                 Object.defineProperty(object, property, properties[property]);
             }
-        }
+        });
         return object;
     };
 }
--- a/tests/spec/s-object.js
+++ b/tests/spec/s-object.js
@@ -233,4 +233,17 @@ describe('Object', function () {
             }
         });
     });
+
+    describe('Object.defineProperties', function () {
+        it('should define the constructor property', function () {
+            var target = {};
+            var newProperties = {
+                constructor: {
+                  value: 'new constructor'
+                }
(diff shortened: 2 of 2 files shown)


ljharb:
Thanks! This definitely seems like a strictly better approach regardless. Since this particular sham would only apply in pre-ES5 browsers, testing in IE 8 covers down to IE 6, and I'll manually test this in a few older Chrome/Firefox/Safari/Opera versions before merging.

Issue #150, opened 2013-02-06
Object.create fails for IE<9 if document.domain is set


angus-c:
If the main document has domain attribute set then we get a permission denied error trying to access `iframe.contentWindow.Object` (due to same-domain restriction)

Tried to patch it up with very hokey iframe.src code:

```
iframe.src = "javascript:'<script>window.onload=function(){document.write(\\'<script>document.domain=\\\"" + document.domain + "\\\";<\\\\/script>\\');document.close();};<\/script>'";
```

but it didn't do the trick.

This issue is preventing twitter.com from upgrading to latest es5-shim

PR #306, opened 2015-06-06
Support for IE8 when document.domain is set

danjamin:
In relation to issue #150 -- took commit 4f738ac066346 and applied it to **es5-sham**

Basically, the problem is that in IE < 9, when `document.domain` is set, using the **es5-sham** causes a *"Permission denied"* error.  What this fix does, is uses a different approach to the Object.create polyfill when this situation arises.  It leaves the original iframe approach intact for all other cases.

For those of us still supporting **IE8** :weary:, and where `document.domain` is necessary for an API being used via iframes (x-origin reasons), this fix should allow **es5-sham** to be used without error.

### DIFF
--- a/es5-sham.js
+++ b/es5-sham.js
@@ -190,6 +190,64 @@ if (!Object.create) {
                         // the following produces false positives
                         // in Opera Mini => not a reliable check
                         // Object.prototype.__proto__ === null
+
+    // Check for document.domain and active x support
+    // No need to use active x approach when document.domain is not set
+    // see https://github.com/es-shims/es5-shim/issues/150
+    // variation of https://github.com/kitcambridge/es5-shim/commit/4f738ac066346
+    /*global ActiveXObject */
+    var shouldUseActiveX = function shouldUseActiveX() {
+        // return early if document.domain not set
+        if (!document.domain) {
+            return false;
+        }
+
+        try {
+            return !!new ActiveXObject('htmlfile');
+        } catch (exception) {
+            return false;
+        }
+    };
+
+    // This supports IE8 when document.domain is used
+    // see https://github.com/es-shims/es5-shim/issues/150
+    // variation of https://github.com/kitcambridge/es5-shim/commit/4f738ac066346
+    var getEmptyViaActiveX = function getEmptyViaActiveX() {
+        var empty;
+        var xDoc;
+
+        xDoc = new ActiveXObject('htmlfile');
+
+        xDoc.write('<script><\/script>');
+        xDoc.close();
+
+        empty = xDoc.parentWindow.Object.prototype;
+        xDoc = null;
+
+        return empty;
+    };
+
+    // The original implementation using an iframe
+    // before the activex approach was added
+    // see https://github.com/es-shims/es5-shim/issues/150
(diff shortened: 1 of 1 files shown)


es5-sham.js
```diff
@@ -190,6 +190,63 @@ if (!Object.create) {
...
+    var shouldUseActiveX = function shouldUseActiveX() {
+        var SUPPORTS_ACTIVEX;
+
+        try {
+            SUPPORTS_ACTIVEX = !!new ActiveXObject('htmlfile');
+        } catch (exception) {
+            SUPPORTS_ACTIVEX = false;
+        }
+
+        return SUPPORTS_ACTIVEX && document.domain;
```
ljharb: Should this `if (!document.domain) { return false; }` on the first line of the function, to avoid the try/catch when it's not set?

es5-sham.js
```diff
@@ -202,16 +259,12 @@ if (!Object.create) {
...
-            parent.appendChild(iframe);
-            /*eslint-disable no-script-url */
-            iframe.src = 'javascript:';
-            /*eslint-enable no-script-url */
-            var empty = iframe.contentWindow.Object.prototype;
-            parent.removeChild(iframe);
-            iframe = null;
+            // Determine which approach to use
+            // see https://github.com/es-shims/es5-shim/issues/150
+            var empty = shouldUseActiveX() ?
```
ljharb: not a big deal, but I'd prefer this be all on one line :-) multiline ternaries are gnarly.

ljharb:
This is great - I suspect this can't be tested for without making a separate HTML file, since `document.domain` can't be unset once set?

es5-sham.js
danjamin: Agreed :+1:

es5-sham.js
danjamin: Done

danjamin:
@ljharb agreed that testing this specific edge case might be a bit tricky given the IE8 environment and `document.domain` needing to be set

ljharb:
@danjamin It would be super helpful if you could set up a jsfiddle, that used rawgit.com to point to your PR's test file, and set `document.domain` so I can use it to verify in various browser versions :-) if not, i'll wait to merge it til I have time to set that up.

danjamin:
@ljharb I'll give it a shot -- will ping you when it's ready

ljharb:
Thanks!

danjamin:
@ljharb turns out jsfiddle doesn't let you adjust the `document.domain`; so I just created a quick gh-pages for my fork;  you can visit it here http://danjamin.github.io/es5-shim/

it has 2 pages (/broken and /fixed) -- if you visit both in IE8 you will see that the broken one has the "Permission Denied" error where as the fixed one does not error on the `Object.create(null)` invocation.

The only real difference between /broken and /fixed is that one points at es5-shim/master of es5-sham.js and the other at the version in the branch in my fork being pull requested.

ljharb:
Awesome, that's good enough for me! https://github.com/danjamin/es5-shim/commit/c471214c1c8c4f6a1042c2ef53a0e69cf2ac5ad0 confirms it for me in my testing.

danjamin:
Cool -- glad I could help :smiley:

PR #307, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


ljharb:
Awesome, thanks. One less file to update.

Hopefully people aren't using `bower` anyways, since `npm` exists.

kkirsche:
Bower (or who I've talked to) have said to use NPM not them. Bower will slowly die

PR #310, opened 2015-06-17
Increase readability (Date.parse) + fix: Date.parse should be non-enumerable

hypotenuse:


ljharb:
Please split this up into multiple PRs - this diff is not readable, so I can't easily review it.

Also, since `Date.parse` starts out non-enumerable, simply assigning over it preserves the non-enumerability. Can you add some tests that would fail without your changes?

es5-shim.js
```diff
@@ -1010,81 +1010,81 @@ if (!Date.parse || doesNotParseY2KNewYear || acceptsInvalidDates || !supportsExt
...
-
+        
         // Upgrade Date.parse to handle simplified ISO 8601 strings
-        DateShim.parse = function parse(string) {
-            var match = isoDateExpression.exec(string);
-            if (match) {
+        var DateShimParse = function parse(string) {
+            
+            var match, year, month, day, hour, minute, second, millisecond
+               ,isLocalTime, signOffset, hourOffset, minuteOffset, result;
```
ljharb: comma first is never acceptable here, and please either put every variable on the same line (regardless of line length, which doesn't matter), or put each variable on its own line.

es5-shim.js
```diff
@@ -1010,81 +1010,81 @@ if (!Date.parse || doesNotParseY2KNewYear || acceptsInvalidDates || !supportsExt
...
+               
+            if (match = isoDateExpression.exec(string)) {
```
ljharb: Please never conflate assignment with conditional checking. This is two statements and should be on two lines.

es5-shim.js
```diff
@@ -1010,81 +1010,81 @@ if (!Date.parse || doesNotParseY2KNewYear || acceptsInvalidDates || !supportsExt
...
-                    millisecond = Math.floor(Number(match[7] || 0) * 1000),
-                    // When time zone is missed, local offset should be used
-                    // (ES 5.1 bug)
-                    // see https://bugs.ecmascript.org/show_bug.cgi?id=112
-                    isLocalTime = Boolean(match[4] && !match[8]),
-                    signOffset = match[9] === '-' ? 1 : -1,
-                    hourOffset = Number(match[10] || 0),
-                    minuteOffset = Number(match[11] || 0),
-                    result;
+                year    = Number(match[1]);
```
ljharb: Please remove the extra spaces before the "=" - alignment makes diffs noisier and makes code harder to read, and is not acceptable in this repo.

es5-shim.js
```diff
@@ -1010,81 +1010,81 @@ if (!Date.parse || doesNotParseY2KNewYear || acceptsInvalidDates || !supportsExt
...
+                second  = Number(match[6] || 0);
+                
+                millisecond = Math.floor(
+                    Number(match[7] || 0) * 1e3
+                );
+                
+                // When time zone is missed, local offset should be used
+                // (ES 5.1 bug)
+                // see https://bugs.ecmascript.org/show_bug.cgi?id=112
+                isLocalTime = Boolean(match[4] && !match[8]);
```
ljharb: Please use `!!x` instead of `Boolean(x)`, for both performance and clarity.

es5-shim.js
```diff
@@ -1010,81 +1010,81 @@ if (!Date.parse || doesNotParseY2KNewYear || acceptsInvalidDates || !supportsExt
...
-        }, true);
-
+        
         // Upgrade Date.parse to handle simplified ISO 8601 strings
-        DateShim.parse = function parse(string) {
-            var match = isoDateExpression.exec(string);
+        var DateShimParse = function parse(string) {
+            
+            var match, year, month, day, hour, minute, second, millisecond, isLocalTime, signOffset, hourOffset, minuteOffset, result;
+                match = isoDateExpression.exec(string);
```
ljharb: Please do `var match` here, instead of initializing all the variables at the top of the function - that's not a style we insist on.

ljharb:
The Date shims are very complex - this will need to be tested in ES3 and ES5 versions of IE, Safari, Opera, Firefox, Chrome, etc before merging, and definitely will require automated tests to be added.

hypotenuse:
Due to the fact that automated tests are required the commit should be closed then. But at least it worth noting that original ```Date.parse``` method has internal DontEnum attribute which is set to true by default (non-enumerable). Current shim unfortunately has this attribute set to false (enumerable).

ljharb:
I'll double check that - in which browsers? In my understanding, the way JS engines work is that a non-enumerable property that is assigned to using the internal `[[Put]]` operation does not change the `enumerable` descriptor value.

ljharb:
Wow, you are totally correct, thanks! Fix incoming.

PR #312, opened 2015-06-22
Avoid publish dev file

roderickhsiao:
adding `.npmignore` to avoid publish dev files to npm

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+.travis.yml
+/tests
+Makefile


ljharb:
I explicitly want tests published to npm, so that `npm explore es5-shim && npm install && npm test` works. At that point, I don't think travis.yml or the Makefile makes a difference.

roderickhsiao:
Certainly, thanks for your explanation. The first intention of me is to reduce the overall package size, so would like to exclude dev file for publish for most of the package included (expect user will checkout repo to run test).

But you certainly make point that user might want to run test locally by install it via npm.

Thanks

Issue #297, opened 2015-03-19
Object.defineProperty does not fail silently in IE8


schmod:
The documentation for es5-sham seems to imply that `Object.defineProperty()` should fail when setting properties such as `configurable`, `enumerable`, and `writable`.  

However, running the following code throws an error:

```js
Object.defineProperty({foo: 'bar'}, 'foo', {
   configurable: true
})'
```

Looking at the relevant portion of the sham, it seems like we almost always throw an error if getters/setters are unsupported (even if we're not asking for a getter or setter):

```js
Object.defineProperty = function defineProperty(object, property, descriptor) {
    if ((typeof object !== 'object' && typeof object !== 'function') || object === null) {
        throw new TypeError(ERR_NON_OBJECT_TARGET + object);
    }
    if ((typeof descriptor !== 'object' && typeof descriptor !== 'function') || descriptor === null) {
        throw new TypeError(ERR_NON_OBJECT_DESCRIPTOR + descriptor);
    }
    // make a valiant attempt to use the real defineProperty
    // for I8's DOM elements.
    if (definePropertyFallback) {
        try {
            return definePropertyFallback.call(Object, object, property, descriptor);
        } catch (exception) {
            // try the shim if the real one doesn't work
        }
    }

    // If it's a data property.
    if ('value' in descriptor) {
        // fail silently if 'writable', 'enumerable', or 'configurable'
        // are requested but not supported
        /*
        // alternate approach:
        if ( // can't implement these features; allow false but not true
            ('writable' in descriptor && !descriptor.writable) ||
            ('enumerable' in descriptor && !descriptor.enumerable) ||
            ('configurable' in descriptor && !descriptor.configurable)
        ))
            throw new RangeError(
                'This implementation of Object.defineProperty does not support configurable, enumerable, or writable.'
            );
        */

        if (supportsAccessors && (lookupGetter(object, property) || lookupSetter(object, property))) {
            // As accessors are supported only on engines implementing
            // `__proto__` we can safely override `__proto__` while defining
            // a property to make sure that we don't hit an inherited
            // accessor.
            /*eslint-disable no-proto */
            var prototype = object.__proto__;
            object.__proto__ = prototypeOfObject;
            // Deleting a property anyway since getter / setter may be
            // defined on object itself.
            delete object[property];
            object[property] = descriptor.value;
            // Setting original `__proto__` back now.
            object.__proto__ = prototype;
            /*eslint-enable no-proto */
        } else {
[...]

PR #315, opened 2015-06-24
fix define empty descriptor

yiminghe:
It should not throw error for empty descriptor

```js
Object.defineProperty({},'a',{});
```

It is used in intl:  https://github.com/andyearnshaw/Intl.js/blob/1fced3d60d004b9a853e3699cc6f3329096957da/src/core.js#L119

### DIFF
--- a/es5-sham.js
+++ b/es5-sham.js
@@ -413,7 +413,7 @@ if (!Object.defineProperty || definePropertyFallback) {
                 object[property] = descriptor.value;
             }
         } else {
-            if (!supportsAccessors) {
+            if (!supportsAccessors && (('get' in descriptor) || ('set' in descriptor))) {
                 throw new TypeError(ERR_ACCESSORS_NOT_SUPPORTED);
             }
             // If we got that far then getters and setters can be defined !!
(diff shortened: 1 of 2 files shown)


ljharb:
This relates to #297, specifically https://github.com/es-shims/es5-shim/issues/297#issuecomment-94549318

In addition, https://github.com/es-shims/es5-shim/blob/cd80ad124eeba169d1ed69c1cc4a6bdee2ca82a1/es5-sham.js#L340 would need to change as well, preferably setting "value".

I think your change is good, but the question remains, what should happen if the descriptor is *not* empty? ie, if `value` is not provided, `get` and `set` are not provided, but `configurable`, `enumerable`, and `writable` are present? It seems to me like it should throw, whereas with your change, it would silently fail.

yiminghe:
Get and set must throw, because old ie does not support it and the rest of code will fall definitely.

But others does not need to throw, warning is enough.

If you do not want to change, then I have to ask intl ...

ljharb:
OK so since https://github.com/yiminghe/es5-shim/blob/definePropertyFix/es5-sham.js#L383-L395 is the current behavior of the "value" path, it should be the same behavior for the "non-value" path.

If ever those lines become uncommented, they'd be moved up and run in both code paths. So your change shouldn't affect that.

schmod:
Thanks for taking the time to fix this.  This is clearly an improvement over the current status quo :)

ljharb:
I'll release this over the next couple days.

PR #317, opened 2015-07-05
Update bower.json

ajoslin103:
when I installed this package it didn't insert a line into my index.js

I believe changing the main to ./es5-shim.js will correct that

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
     "name": "es5-shim",
-    "main": "es5-shim.js",
+    "main": "./es5-shim.js",
     "repository": {
         "type": "git",
         "url": "git://github.com/es-shims/es5-shim"


ljharb:
What do you mean `insert a line into my index.js`? That's something you have to do manually, bower installing something just puts it into `bower_components`.

ajoslin103:
bower will automatically insert a script inclusion within your bower section (if you have one) of your index file -- if the 'main' param in the bower.json file includes the path to the script file (at least that's how it works for me)

ljharb:
I don't see anything about that in https://github.com/bower/bower.json-spec#main - and none of their examples use leading `.` in the filenames

ajoslin103:
let me go find something - I use bower a lot and it always inserts on install and removes on uninstall

maybe I'm a little mistaken, but it's been working like that for me for a long time

ljharb:
I'm fine to change it, but since I don't use bower (because npm exists) I don't want to risk breaking it for people who use it. Documentation would help :-)

ajoslin103:
I can certainly understand that - I just checked a few packages that installed ok and they don't have leading . either -- still looking

ajoslin103:
my mistake -- it's wiredep that does the task

http://www.ibm.com/developerworks/websphere/library/techarticles/1503_maynard/1503_maynard.html

and it did work for es5-shim -- it's just that it put it somewhere I wasn't expecting

sorry for the distraction

:)

ljharb:
Thanks for researching it!

ajoslin103:
np, I should have checked before hollering !

I hope wiredep helps you, looked like a setup for npm in that tutorial

PR #320, opened 2015-07-15
Add a typeof check to Object.getPrototypeOf. [closes #319, #320]

### LINKED ISSUE #319: Object.getPrototypeOf constructor check.
I noticed the current `Object.getPrototypeOf` check:
```js
if (object.constructor)
```
but the constructor can really be any value so maybe a more thorough check like
```js
if (typeof object.constructor === 'function')
```
would be better.

jdalton:
See #319

### DIFF
--- a/es5-sham.js
+++ b/es5-sham.js
@@ -33,6 +33,7 @@ var call = Function.prototype.call;
 var prototypeOfObject = Object.prototype;
 var owns = call.bind(prototypeOfObject.hasOwnProperty);
 var propertyIsEnumerable = call.bind(prototypeOfObject.propertyIsEnumerable);
+var toStr = call.bind(prototypeOfObject.toString);
 
 // If JS engine supports accessors creating shortcuts.
 var defineGetter;
@@ -65,11 +66,10 @@ if (!Object.getPrototypeOf) {
         /* eslint-enable no-proto */
         if (proto || proto === null) {
             return proto;
-        } else if (object.constructor) {
+        } else if (toStr(object.constructor) === '[object Function]') {
             return object.constructor.prototype;
-        } else {
-            return prototypeOfObject;
         }
+        return prototypeOfObject;
     };
 }
 
--- a/tests/spec/s-object.js
+++ b/tests/spec/s-object.js
@@ -260,6 +260,38 @@ describe('Object', function () {
         });
     });
 
+    describe('Object.getPrototypeOf', function () {
+        it('should return the [[Prototype]] of an object', function () {
+            var Foo = function () {};
+
+            var proto = Object.getPrototypeOf(new Foo());
+
+            expect(proto).toBe(Foo.prototype);
+        });
+
+        it('should shamone to the `Object.prototype` if `object.constructor` is not a function', function () {
+            var Foo = function () {};
+            Foo.prototype.constructor = 1;
+
+            var proto = Object.getPrototypeOf(new Foo()),
+                fnToString = Function.prototype.toString;
+
+            if (fnToString.call(Object.getPrototypeOf).indexOf('[native code]') < 0) {
+                expect(proto).toBe(Object.prototype);
+            } else {
+                expect(proto).toBe(Foo.prototype);
+            }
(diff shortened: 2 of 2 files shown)


jdalton:
Travis test fails due to npm.

es5-sham.js
```diff
@@ -65,7 +65,7 @@ if (!Object.getPrototypeOf) {
         /* eslint-enable no-proto */
         if (proto || proto === null) {
             return proto;
-        } else if (object.constructor) {
+        } else if (typeof object.constructor === 'function') {
```
ljharb: Will this not return a false positive with regexes in some older engines?

I think we need an `Object#toString` check here instead.

ljharb:
Can we add a test with `constructor` set to a non-function, to prevent a regression?

jdalton:
Ah nice!

jdalton:
There are currently no tests for `Object.getPrototypeOf`.

es5-sham.js
jdalton: Correct. I saw `typeof` checks in other places so wasn't sure your chance on it.

jdalton:
Is there a way to detect if the test is running in enviros where the built-in exists and isn't shimmed?

ljharb:
Any tests should run only when the builtin exists, and shouldn't care whether it's shimmed or not, ideally.

jdalton:
Yap, gotta test the sham behavior which means different behavior with and without the native built-in.

tests/spec/s-object.js
```diff
@@ -260,6 +260,38 @@ describe('Object', function () {
...
+        });
+
+        it('should shamone to the `Object.prototype` if `object.constructor` is not a function', function () {
+            function Foo() {}
+            Foo.prototype.constructor = 1;
+
+            var proto = Object.getPrototypeOf(new Foo),
+                fnToString = Function.prototype.toString;
+
+            if (fnToString.call(Object.getPrototypeOf).indexOf('[native code]') < 0) {
```
ljharb: While this works, it makes me cringe - why is it that native code will return `Object.prototype` but the sham will return `Foo.prototype`? I'd rather the sham threw, than return the wrong value.

tests/spec/s-object.js
```diff
@@ -260,6 +260,38 @@ describe('Object', function () {
         });
     });
 
+    describe('Object.getPrototypeOf', function () {
+        it('should return the [[Prototype]] of an object', function () {
+            function Foo() {}
```
ljharb: linting rules forbid function declarations - `var Foo = function Foo() {};` would work here.

tests/spec/s-object.js
```diff
@@ -260,6 +260,38 @@ describe('Object', function () {
...
+
+            var proto = Object.getPrototypeOf(new Foo);
```
ljharb: `new Foo` needs to be `new Foo()`

tests/spec/s-object.js
```diff
@@ -260,6 +260,38 @@ describe('Object', function () {
...
+            if (fnToString.call(Object.getPrototypeOf).indexOf('[native code]') < 0) {
+                expect(proto).toBe(Object.prototype);
+            } else {
+                expect(proto).toBe(Foo.prototype);
+            }
+        });
+
+        it('should throw error for non object', function () {
+            try {
+                expect(Object.getPropertyOf(1)).toBeUndefined();
```
ljharb: s/`getPropertyOf`/`getPrototypeOf`, otherwise you'll get "undefined is not a function" :-)

ljharb:
I haven't looked at the sham code in awhile - is the difference in behavior because of unshimmable semantics?

ljharb:
Would you classify this as a breaking change in the sham? or as a bug fix?

If it'll be a patch, let's rebase this and I'll merge and release it. If it's a breaking change, I'll be a bit more cautious.

jdalton:
It's more a bug fix because before the constructor could have been a primitive which isn't the intent of the path.

Issue #301, opened 2015-05-01
Object.create(null) incorrectly sets __proto__ in IE 8


dgreensp:
One of the great things about `Object.create(null)` is that `'__proto__' in Object.create(null)` is false.  This is true in modern browsers and IE 9.

You can easily fix this for IE 8 in the sham -- just don't set `__proto__`!

ExE-Boss:
#321 has been merged for 4 years now, I think this should've been closed by now.

PR #321, opened 2015-07-20
Fix Object.create sham to not set __proto__

dgreensp:
Background:
Object.create(null) is great, because it creates objects whose key
space is not polluted by names like "toString" and "constructor".
In addition, in modern browsers, objects created with
Object.create(null) do not even have a `__proto__` property!  This means
`'__proto__' in x` is false and `x.__proto__ === undefined` when
`x = Object.create(null)`.  Browsers extend this courtesy even further;
if you even set `x.__proto__ = null` at any time, the `__proto__` property
of x loses its special meaning and becomes safe to assign without
actually changing the internal prototype pointer of x.

The Object.create sham uses a very clever technique to create "nullary"
objects in browsers that don't support `__proto__` (like IE <11).
However, it then turns around and sets a property named `__proto__` on the
object!  Note that the code has already determined that this property
has no special meaning.

This commit simply removes the assignment to `__proto__`, so that
`'__proto__' in x` is false and `x.__proto__ === undefined`, as it
should be.

See live correctness test at: http://es5-sham-create-null.meteor.com

See #301.

### DIFF
--- a/tests/spec/s-object.js
+++ b/tests/spec/s-object.js
@@ -312,6 +312,12 @@ describe('Object', function () {
                 expect(err).toEqual(jasmine.any(TypeError));
             }
         });
+
+        it('should return null on Object.create(null)', function () {
+            var obj = Object.create(null);
+
+            expect(Object.getPrototypeOf(obj)).toBe(null);
+        });
     });
 
     describe('Object.defineProperties', function () {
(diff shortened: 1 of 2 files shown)


ljharb:
This was added in commit c608bb2c by @Benvie - I'd love to get his thoughts before merging this change in.

That said, your reasoning and explanation is in-depth and seems sound - but I want to be cautious :-)

ljharb:
This is also a duplicate of #290.

michaelficarra:
For the record, I still think we should make this change. But we should close this PR in favour of #290.

ljharb:
Agreed - they're the same diff, and this is the later one.

dgreensp:
@ljharb, see if you agree this is ready to go!

I learned from writing tests (thanks Travis!) that in Node 0.10, you can actually see the `__proto__` on an `Object.create(null)` object (as a non-enumerable property set to null), while in Node 0.12 it is gone.  So I relaxed my tests a bit.

tests/spec/s-object.js
```diff
@@ -312,6 +312,12 @@ describe('Object', function () {
                 expect(err).toEqual(jasmine.any(TypeError));
             }
         });
+
+        it('should return null on Object.create(null)', function () {
+            var obj = Object.create(null);
+
+            expect(Object.getPrototypeOf(obj) === null).toBe(true);
```
ljharb: `expect(Object.getPrototypeOf(obj)).toBe(null)` will have a nicer failure message

ljharb:
Thanks, this looks good - I'll need some time to do browser tests, but otherwise it's basically good to go :-)

dgreensp:
Great, thanks!  Fixed the test to say `.toBe(null)` and rebased.

tests/spec/s-object.js
dgreensp: Fixed

PR #323, opened 2015-08-04
Add Date.parse checking

hypotenuse:
Please add Date.parse checking. This is reasonable because Date.parse may not exist in the ```if``` statement below :point_down: 
```js
if(!Date.parse || ...
```

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -1066,7 +1066,7 @@ if (!dateToJSONIsSupported) {
 var supportsExtendedYears = Date.parse('+033658-09-27T01:46:40.000Z') === 1e15;
 var acceptsInvalidDates = !isNaN(Date.parse('2012-04-04T24:00:00.500Z')) || !isNaN(Date.parse('2012-11-31T23:59:59.000Z')) || !isNaN(Date.parse('2012-12-31T23:59:60.000Z'));
 var doesNotParseY2KNewYear = isNaN(Date.parse('2000-01-01T00:00:00.000Z'));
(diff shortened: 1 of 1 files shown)


ljharb:
In which browsers/engines does `Date.parse` not exist?

hypotenuse:
Why we use ```!Date.parse``` in the ```if``` statement then? Probably because ```Date.parse``` may not exist. In this case it would be better to omit ```!Date.parse``` from the ```if``` statement. That's it.
Otherwise, I don't know how we can call the ```Date.parse``` method assuming ```!Date.parse``` in the ```if``` statement.

ljharb:
So, this PR is based on reading the code and assuming there's a case it would fail - but not actually having it fail?

I'd prefer to remove the `!Date.parse` check from the `if`, absent actual evidence that a JS engine doesn't ship with it. Since it's part of the initial spec, I don't see how that's possible.

hypotenuse:
This PR is based on reading the code and a question why !Date.parse is in the if statement.

ljharb:
Thank you for pointing out the dissonance :-) I think the solution here is to remove the check in the if statement. Do you want to update this PR to do that?

hypotenuse:
Perfect! Let's do this. New PR is coming.

ljharb:
If possible, please update this one in-place, rebase, and force push :-)

ljharb:
Can you rebase these two commits down to one, with a meaningful commit message?

hypotenuse:
ok

ljharb:
Thanks!

PR #324, opened 2015-08-13
Remove unnecessary Date constructor property

dgreensp:
I question whether this code does anything useful given the presence of `Date.prototype.constructor`.  It actually broke a unit test in our of libraries, for obscure reasons, because it causes `(new Date).hasOwnProperty("constructor")` to return true.

I wonder what sort of mix-ups it is supposed to prevent?

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -1093,8 +1093,6 @@ if (!Date.parse || doesNotParseY2KNewYear || acceptsInvalidDates || !supportsExt
             } else {
                 date = NativeDate.apply(this, arguments);
             }
-            // Prevent mixups with unfixed Date object
-            defineProperties(date, { constructor: DateShim }, true);
             return date;
         };
 


dgreensp:
Oh I see, this code causes `(new Date).constructor` to be equal to `Date` (where in all cases `Date` is actually `DateShim`).

ljharb:
Curious, why is your library checking if a Date has an own "constructor" property or not?

dgreensp:
It's in argument-handling code that is trying to detect whether an argument is an object literal or "some other kind of object."  The library provides a syntax for specifying an HTML template AST, so you can write `DIV("Hello")` or `DIV({...attrs...}, "Hello, ", SPAN(...))`, and because it's an extensible AST, the code tries to detect `{...attrs...}` in a way that won't fire on custom objects that are "instances of a class."  The "some other kind of object" check involves looking for a non-`Object` `constructor`, but in order to not be fooled by an object literal like `{constructor: ...}` it also checked that `constructor` is not an own property.  So, to use custom objects in this extensible AST, you just have to make sure they have a `constructor` property that's not `Object` and isn't an own property.

In writing the unit tests, I happened to use `new Date` as an example of a non-object-literal.  But when `new Date` got an own property named `constructor` from es5-shim, my code no longer recognized it as a valid custom object.

It was really my code that was at fault (for applying a pretty poor criterion in the argument-handling code, and then making assumptions about Date objects in the tests).  A much better test for whether an object is an "instance of a class" than whether `constructor` is an own property is whether the object is `instanceof` its `.constructor` (which is not `Object`).

ljharb:
If you want to detect a date object, you need all the code in https://www.npmjs.com/package/is-date-object :-) (or you could just use that module). `instanceof` is unreliable, as-is `Object#toString.call` in ES6.

dgreensp:
It's not related to my case, but that's interesting.  Calling a `Date.prototype` method throws an error in case of a non-Date, but succeeds on a cross-realm Date, in every browser?

ljharb:
Yes, per spec. In an environment without `Symbol.toStringTag`, `Object.prototype.toString.call(foo) === '[object Date]';` is reliable, but in a newer environment, you need to try/catch one of the prototype methods that throws when there's no internal slot.

Issue #304, opened 2015-05-22
Object.prototype.toString for pre ES5.1 browers


pygy:
Would you be interested in this?

Pasting from [here](https://github.com/Financial-Times/polyfill-service/pull/401):

----

Object.prototype.toString() is sometimes used as a saner type checker. For example:

```JavaScript
type = {}.toString;

type.call("foo") === "[object String]"
type.call(new String("foo")) === "[object String]"

type.call([]) === "[object Array]"
// etc...
```
It has a lot of advantages over typeof and instanceof for the above types.

In modernish Browsers (ES 5.1) the following also holds:

```JavaScipr
type.call(null) === "[object Null]"
type.call(undefined) === "[object Undefined]"
```

`IE < 9`, `FF < 4` and maybe other browsers however return `'[object Object]'` for these types.

----

Implementation:

```JavaScript

if (Object.prototype.toString.call(null) !== "[object Null]" 
 || Object.prototype.toString.call(undefined) !== "[object Undefined]"
) {
    var Object_prototype_toString = Object.prototype.toString
    Object.prototype.toString = function() {
        if (this === null) return "[object Null]";
        if (this === undefined) return "[object Undefined]";
        return Object_prototype_toString.call(this);
    }
}
```

kitcambridge:
This is a tricky one for two reasons. First, because `this` (pardon the pun :smile_cat:) always defaults to the global scope—`window` in the browser, `global` in Node, `self` in a worker—if set to `null` or `undefined`. So, in ES 3, `{}.toString.call(null) == {}.toString.call(window)`.

This is the default behavior for `call` and `apply` in ES 5, too:

```js
(function(){ return function() { return this; }.call(null) }()) == window;
// => true
(function(){ return function() { return this; }.call(undefined) }()) == window;
// => true
```

Apart from `{}.toString`, the other exception is strict mode:

```js
(function(){ 'use strict'; return function() { return this; }.call(null) }()) === null;
// => true
(function(){ 'use strict'; return function() { return this; }.call(undefined) }()) === undefined;
// => true
```

...But strict mode is ES 5-only, so those semantics can't be replicated in ES 3.

The second consequence of overriding `Object.prototype.toString` is that the `"toString"` property will now be enumerated in every `for...in` loop, since ES 3 doesn't let user code mark properties as non-enumerable. It's possible to work around that with `hasOwnProperty`, but not all code does this, and sometimes, you want to enumerate all inherited properties.

[...]

Issue #99, opened 2012-02-01
Fix Function.apply to work with generic array-like object instead of an array


termi:
Can we fix that in IE < 9 ?

I fix it (**outdated**)https://github.com/termi/es5-shim/blob/master/es5-shim.js#L198
But I am not sure, what my solution quite well. Can you check it out?

P.S. In my own lib I use `Array.from` instead of:

```
args = toObject(args);
var i = -1, arr = [], length = args.length >>> 0;

while (++i < length) {
    if (i in args) {
        arr.push(args[i])
    }
}
```

ljharb:
This also affects OmniWeb v5.11.2

Xotic750:
This is a subject that I have been looking at for some of my own code and I thought I'd share my investigation code with you as it may be of some help (maybe). Sorry about the size of it, I don't know how to make the code block scrollable. :)

Available on [jsFiddle](http://jsfiddle.net/Xotic750/rg3j8x4t/)

```javascript
/*global Date, Function, Math, Object, String */

/*properties
    abs, apply, call, floor, length, log, max, min, prototype, random
*/

/*jslint browser: true, devel: true, node: true, sloppy: true*/

(function (global) {
    /**
     * Returns an arguments object of the arguments supplied.
     *
     * @private
     * @function $returnArgs
     * @param {...*} [varArgs]
     * @return {Arguments}
     */
    function $returnArgs() {
        return arguments;
    }

    (function (call, apply) {
        'use strict';

        /*jslint vars: true */

        /**
         * @name forceCallShim
         * @type {boolean}
         */
        var forceCallShim = false;

        /**
         * @name forceCallPatch
         * @type {boolean}
         */
        var forceCallPatch = false;

        /**
         * @name forceApplyShim
         * @type {boolean}
         */
        var forceApplyShim = false;

        /**
         * @name forceApplyPatch
         * @type {boolean}
         */
        var forceApplyPatch = false;

        /**
         * @name MAX_SAFE_INTEGER
         * @type {number}
         * @const
         * @default 9007199254740991
         */
        var MAX_SAFE_INTEGER = 9007199254740991;

        /**
         * Shortcut
         * The Object constructor creates an object wrapper.
         * Keeps jslint happy
         *
         * @private
         * @function $Object
         * @param {*} inputArg
         * @return {Object}
         */
        var $Object = Object;

        /**
         * Shortcut
         * The String constructor creates an object wrapper.
         * Keeps jslint happy
         *
         * @private
[...]

PR #345, opened 2015-11-15
Fix `Function.prototype.call` and tests.

Xotic750:
Ref: https://github.com/es-shims/es5-shim/issues/304

This is not necessarily ready to pull, but I want for the tests to be run.

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -56,6 +56,7 @@ var array_push = ArrayPrototype.push;
 var array_unshift = ArrayPrototype.unshift;
 var array_concat = ArrayPrototype.concat;
 var call = FunctionPrototype.call;
+var apply = FunctionPrototype.apply;
 var max = Math.max;
 var min = Math.min;
 
@@ -163,7 +164,6 @@ var ES = {
     // http://es5.github.com/#x9.9
     /* replaceable with https://npmjs.com/package/es-abstract ES5.ToObject */
     ToObject: function (o) {
-        /* jshint eqnull: true */
         if (o == null) { // this matches both null and undefined
             throw new TypeError("can't convert " + o + ' to object');
         }
@@ -180,6 +180,68 @@ var ES = {
 // Function
 // ========
 //
+
+// Tests for inconsistent or buggy `[[Class]]` strings.
+/* eslint-disable no-useless-call */
+var hasToStringTagBasicBug = to_string.call() !== '[object Undefined]' || to_string.call(null) !== '[object Null]';
+/* eslint-enable no-useless-call */
+var hasToStringTagLegacyArguments = to_string.call(arguments) !== '[object Arguments]';
+var hasToStringTagInconsistency = hasToStringTagBasicBug || hasToStringTagLegacyArguments;
+// Others that could be fixed:
+// Older ES3 native functions like `alert` return `[object Object]`.
+// Inconsistent `[[Class]]` strings for `window` or `global`.
+
+if (hasToStringTagInconsistency) {
+    // To prevent recursion when `Function#call` is patched. Robustness.
+    call.call = call;
+}
+
+if (hasToStringTagLegacyArguments) {
+    // This function is for use within `call` only.
+    // To avoid any possibility of `call` recursion we use original `hasOwnProperty`.
+    var isDuckTypeArguments = (function (hasOwnProperty) {
+      return function (value) {
+          return value != null && // Checks `null` or `undefined`.
+              typeof value === 'object' &&
+              call.call(hasOwnProperty, value, 'length') &&
+              typeof value.length === 'number' &&
+              value.length >= 0 &&
+              !call.call(hasOwnProperty, value, 'arguments') &&
+              call.call(hasOwnProperty, value, 'callee');
+              // Using `isCallable` here is dangerous as it can cause recursion.
+              // Unless modified to use `call.call`.
+              // A typeof check would be ok if necessary.
+        };
+    }(ObjectPrototype.hasOwnProperty));
+}
+
+// ES-5 15.3.4.4
+// http://es5.github.io/#x15.3.4.4
+// The call() method calls a function with a given this value and arguments
+// provided individually.
+defineProperties(FunctionPrototype, {
+  call: function (thisArg) {
+      // If `this` is `Object#toString`.
+      if (this === to_string) {
+          // Add whatever fixes for getting `[[Class]]` strings here.
+          if (thisArg === null) {
+              return '[object Null]';
+          }
+          if (typeof thisArg === 'undefined') {
+              return '[object Undefined]';
+          }
+          if (hasToStringTagLegacyArguments && isDuckTypeArguments(thisArg)) {
+              return '[object Arguments]';
+          }
+          return call.call(to_string, thisArg);
+      }
+      // All other calls.
+      // `array_slice` is safe to use here, no known issue at present.
+      // Could probably use eval/Function here to spread the arguments, then use
+      // `call.call` to avoid `apply` altogether.
+      return this.apply(thisArg, call.call(array_slice, arguments, 1));
+  }
+}, hasToStringTagInconsistency);
 
 // ES-5 15.3.4.5
 // http://es5.github.com/#x15.3.4.5
@@ -321,7 +383,7 @@ defineProperties(FunctionPrototype, {
 });
 
 // _Please note: Shortcuts are defined after `Function.prototype.bind` as we
-// us it in defining shortcuts.
+// use it in defining shortcuts.
 var owns = call.bind(ObjectPrototype.hasOwnProperty);
 var toStr = call.bind(ObjectPrototype.toString);
 var strSlice = call.bind(StringPrototype.slice);
--- a/tests/spec/s-function.js
+++ b/tests/spec/s-function.js
@@ -1,7 +1,84 @@
-/* global describe, it, expect, beforeEach */
+/* global describe, it, xit,  expect, beforeEach */
+var hasStrictMode = (function () {
+    'use strict';
+
+    return !this;
+}());
+var ifHasStrictIt = hasStrictMode ? it : xit;
+var global = Function('return this')();
 
 describe('Function', function () {
-    'use strict';
+    describe('#call()', function () {
+        it('should pass correct arguments', function () {
+            // https://github.com/es-shims/es5-shim/pull/345#discussion_r44878754
+            var result;
+            var testFn = function () {
+                return Array.prototype.slice.call(arguments);
+            };
+            var argsExpected = [null, '1', 1, true, testFn];
+            /* eslint-disable no-useless-call */
+            result = testFn.call(undefined, null, '1', 1, true, testFn);
+            expect(result).toEqual(argsExpected);
+            result = testFn.call(null, null, '1', 1, true, testFn);
+            expect(result).toEqual(argsExpected);
+            /* eslint-enable no-useless-call */
+            result = testFn.call('a', null, '1', 1, true, testFn);
+            expect(result).toEqual(argsExpected);
+            result = testFn.call(1, null, '1', 1, true, testFn);
+            expect(result).toEqual(argsExpected);
+            result = testFn.call(true, null, '1', 1, true, testFn);
+            expect(result).toEqual(argsExpected);
+            result = testFn.call(testFn, null, '1', 1, true, testFn);
+            expect(result).toEqual(argsExpected);
+            result = testFn.call(new Date(), null, '1', 1, true, testFn);
+            expect(result).toEqual(argsExpected);
+        });
+        // https://github.com/es-shims/es5-shim/pull/345#discussion_r44878771
+        ifHasStrictIt('should have correct context in strict mode', function () {
+            'use strict';
+
+            var subject;
+            var testFn = function () {
+                return this;
+            };
+            expect(testFn.call()).toBe(undefined);
+            /* eslint-disable no-useless-call */
+            expect(testFn.call(undefined)).toBe(undefined);
+            expect(testFn.call(null)).toBe(null);
+            /* eslint-enable no-useless-call */
+            expect(testFn.call('a')).toBe('a');
+            expect(testFn.call(1)).toBe(1);
+            expect(testFn.call(true)).toBe(true);
+            expect(testFn.call(testFn)).toBe(testFn);
+            subject = new Date();
+            expect(testFn.call(subject)).toBe(subject);
+        });
+        it('should have correct context in non-strict mode', function () {
+            var result;
+            var subject;
+            var testFn = function () {
+                return this;
+            };
+
+            expect(testFn.call()).toBe(global);
+            /* eslint-disable no-useless-call */
+            expect(testFn.call(undefined)).toBe(global);
+            expect(testFn.call(null)).toBe(global);
+            /* eslint-enable no-useless-call */
+            result = testFn.call('a');
+            expect(typeof result).toBe('object');
+            expect(String(result)).toBe('a');
+            result = testFn.call(1);
+            expect(typeof result).toBe('object');
+            expect(Number(result)).toBe(1);
+            result = testFn.call(true);
+            expect(typeof result).toBe('object');
+            expect(Boolean(result)).toBe(true);
+            expect(testFn.call(testFn)).toBe(testFn);
+            subject = new Date();
+            expect(testFn.call(subject)).toBe(subject);
+        });
+    });
 
     describe('#apply()', function () {
         it('works with arraylike objects', function () {
--- a/tests/spec/s-object.js
+++ b/tests/spec/s-object.js
@@ -1,4 +1,6 @@
-/* global describe, it, xit, expect, beforeEach, jasmine, window */
+/* global describe, it, xit, expect, beforeEach, jasmine, window,
+  ArrayBuffer, Float32Array, Float64Array, Int8Array, Int16Array,
+  Int32Array, Uint8Array, Uint8ClampedArray, Uint16Array, Uint32Array */
 
 var ifWindowIt = typeof window === 'undefined' ? xit : it;
 var extensionsPreventible = typeof Object.preventExtensions === 'function' && (function () {
@@ -22,7 +24,14 @@ var canFreeze = typeof Object.freeze === 'function' && (function () {
     return obj.a !== 3;
 }());
 var ifCanFreezeIt = canFreeze ? it : xit;
-
+var toStr = Object.prototype.toString;
+var noop = function () {};
+var hasIteratorTag = typeof Symbol === 'function' && typeof Symbol.iterator === 'symbol';
+var ifHasIteratorTag = hasIteratorTag ? it : xit;
+var hasArrayBuffer = typeof ArrayBuffer === 'function';
+var ifHasArrayBuffer = hasArrayBuffer ? it : xit;
+var hasUint8ClampedArray = typeof Uint8ClampedArray === 'function';
+var ifHasUint8ClampedArray = hasUint8ClampedArray ? it : xit;
 describe('Object', function () {
     'use strict';
 
@@ -354,6 +363,61 @@ describe('Object', function () {
             expect(protoIsEnumerable).toBe(false);
 
             expect(obj instanceof Object).toBe(false);
+        });
+    });
+
+    describe('#toString', function () {
+        it('basic', function () {
+            expect(toStr.call()).toBe('[object Undefined]');
+            /* eslint-disable no-useless-call */
+            expect(toStr.call(undefined)).toBe('[object Undefined]');
+            expect(toStr.call(null)).toBe('[object Null]');
+            /* eslint-enable no-useless-call */
+            expect(toStr.call(1)).toBe('[object Number]');
+            expect(toStr.call(true)).toBe('[object Boolean]');
+            expect(toStr.call('x')).toBe('[object String]');
+            expect(toStr.call([1, 2, 3])).toBe('[object Array]');
+            expect(toStr.call(arguments)).toBe('[object Arguments]');
+            expect(toStr.call({})).toBe('[object Object]');
+            expect(toStr.call(noop)).toBe('[object Function]');
+            expect(toStr.call(new RegExp('c'))).toBe('[object RegExp]');
+            expect(toStr.call(new Date())).toBe('[object Date]');
+            expect(toStr.call(new Error('x'))).toBe('[object Error]');
+        });
+        ifHasArrayBuffer('Typed Arrays', function () {
+            var buffer = new ArrayBuffer(8);
+            expect(toStr.call(buffer)).toBe('[object ArrayBuffer]');
+            expect(toStr.call(new Float32Array(buffer))).toBe('[object Float32Array]');
+            expect(toStr.call(new Float64Array(buffer))).toBe('[object Float64Array]');
+            expect(toStr.call(new Int8Array(buffer))).toBe('[object Int8Array]');
+            expect(toStr.call(new Int16Array(buffer))).toBe('[object Int16Array]');
+            expect(toStr.call(new Int32Array(buffer))).toBe('[object Int32Array]');
+            expect(toStr.call(new Uint8Array(buffer))).toBe('[object Uint8Array]');
+            expect(toStr.call(new Uint16Array(buffer))).toBe('[object Uint16Array]');
+            expect(toStr.call(new Uint32Array(buffer))).toBe('[object Uint32Array]');
+        });
+        ifHasUint8ClampedArray('Uint8ClampedArray', function () {
+            var buffer = new ArrayBuffer(8);
+            expect(toStr.call(new Uint32Array(buffer))).toBe('[object Uint32Array]');
+        });
+        ifHasIteratorTag('Symbol.iterator', function () {
+            expect(toStr.call(Symbol.iterator)).toBe('[object Symbol]');
+        });
+        // https://github.com/es-shims/es5-shim/pull/345#discussion_r44878834
+        it('prototypes', function () {
+            expect(toStr.call(Object.prototype)).toBe('[object Object]');
+            expect(toStr.call(Array.prototype)).toBe('[object Array]');
+            expect(toStr.call(Boolean.prototype)).toBe('[object Boolean]');
+            expect(toStr.call(Function.prototype)).toBe('[object Function]');
+        });
+        // In ES6, many prototype objects stop being instances of themselves,
+        // and instead would return '[object Object]'.
+        xit('prototypes', function () {
+            expect(toStr.call(Number.prototype)).toBe('[object Number]');
+            expect(toStr.call(String.prototype)).toBe('[object String]');
+            expect(toStr.call(Error.prototype)).toBe('[object Error]');
+            expect(toStr.call(Date.prototype)).toBe('[object Date]');
+            expect(toStr.call(RegExp.prototype)).toBe('[object RegExp]');
         });
     });
 });
--- a/tests/spec/s-array.js
+++ b/tests/spec/s-array.js
@@ -9,6 +9,7 @@ var hasStrictMode = (function () {
 
    return !this;
 }());
+var ifHasStrictIt = hasStrictMode ? it : xit;
 
 describe('Array', function () {
     var testSubject;
@@ -123,17 +124,15 @@ describe('Array', function () {
             expect(toStr.call(listArg)).toBe('[object String]');
         });
 
-        if (hasStrictMode) {
-            it('does not autobox the content in strict mode', function () {
-                var context;
-                [1].forEach(function () {
-                    'use strict';
+        ifHasStrictIt('does not autobox the content in strict mode', function () {
+            var context;
+            [1].forEach(function () {
+                'use strict';
 
-                    context = this;
-                }, 'x');
-                expect(typeof context).toBe('string');
-            });
-        }
+                context = this;
+            }, 'x');
+            expect(typeof context).toBe('string');
+        });
     });
     describe('#some()', function () {
         var actual, expected, numberOfRuns;


Xotic750:
All seems well so far on Node. Is there somewhere to see the automated browser tests (if there are any)?

es5-shim.js
```diff
@@ -6,12 +6,15 @@
 
 // vim: ts=4 sts=4 sw=4 expandtab
 
-// Add semicolon to prevent IIFE from being passed as argument to concatenated code.
-;
+// Added to override editor options
+/*jshint maxparams:false, maxdepth:false, maxstatements:false, maxcomplexity:false,
```
ljharb: please revert this - we don't use jshint in this repo.

es5-shim.js
```diff
@@ -64,7 +68,6 @@ var to_string = ObjectPrototype.toString;
 
 var hasToStringTag = typeof Symbol === 'function' && typeof Symbol.toStringTag === 'symbol';
 var isCallable; /* inlined from https://npmjs.com/is-callable */ var fnToStr = Function.prototype.toString, tryFunctionObject = function tryFunctionObject(value) { try { fnToStr.call(value); return true; } catch (e) { return false; } }, fnClass = '[object Function]', genClass = '[object GeneratorFunction]'; isCallable = function isCallable(value) { if (typeof value !== 'function') { return false; } if (hasToStringTag) { return tryFunctionObject(value); } var strClass = to_string.call(value); return strClass === fnClass || strClass === genClass; };
-var isRegex; /* inlined from https://npmjs.com/is-regex */ var regexExec = RegExp.prototype.exec, tryRegexExec = function tryRegexExec(value) { try { regexExec.call(value); return true; } catch (e) { return false; } }, regexClass = '[object RegExp]'; isRegex = function isRegex(value) { if (typeof value !== 'object') { return false; } return hasToStringTag ? tryRegexExec(value) : to_string.call(value) === regexClass; };
```
ljharb: please revert this as well - this is required because `Object#toString` is not reliable in ES6.

es5-shim.js
```diff
@@ -176,6 +179,36 @@ var ES = {
     }
 };
 
+/**
+ * An implementation of `slice`. Primarily for use with `arguments`.
```
ljharb: we already have `sliceArgs` internally, will this not work?

es5-shim.js
Xotic750: I know, it was just while in my editor, I will remove stuff that has nothing to do with the fix.

es5-shim.js
Xotic750: No problem, just wanted to for my code path for now.

es5-shim.js
Xotic750: Aha, I didn't notice it.

Xotic750: Can't find `sliceArgs`? Need it instead of `Array#slice` because I believe that in some of the environments that these fixes are for, it doesn't slice arguments.

ljharb: oh oops, i'm referring to something in the es6-shim :-) https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L1252-L1263

es5-shim.js
```diff
@@ -176,6 +177,36 @@ var ES = {
...
+ *
+ * @private
```
ljharb: also please remove the jsdoc, i don't permit it in any projects i maintain

es5-shim.js
```diff
@@ -176,6 +177,36 @@ var ES = {
...
+ * @param {Array} array The array to slice.
+ * @param {number} [start=0] The start position.
+ * @param {number} [end=array.length] The end position.
+ * @returns {Array} Returns the slice of `array`.
+ */
+var baseSlice = function (array, start, end) {
+  /* jshint eqnull: true */
```
ljharb: please remove the jshint directive

es5-shim.js
```diff
@@ -329,6 +360,97 @@ var strSplit = call.bind(StringPrototype.split);
 var strIndexOf = call.bind(StringPrototype.indexOf);
 var push = call.bind(array_push);
 
+// ES-5 15.3.4.4
+// http://es5.github.io/#x15.3.4.4
+var hasToStringTagBasicBug = !(toStr() === '[object Undefined]' && toStr(null) === '[object Null]');
```
ljharb: instead of !(a && b)` could this be `!a || !b`?

es5-shim.js
```diff
@@ -329,6 +360,97 @@ var strSplit = call.bind(StringPrototype.split);
...
+var hasToStringTagArgumentsBug = toStr(function () {
+  return arguments;
+}) !== '[object Arguments]';
```
ljharb: this isn't a bug so much as a legacy arguments object - which is already checked [here](https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L932-L945)

es5-shim.js
Xotic750: Ok

es5-shim.js
Xotic750: Will remove all of them

es5-shim.js
Xotic750: Indeed

es5-shim.js
Xotic750: Yep, but it's different in different environments. All ES5 environments give `[object Arguments]`, whether in the spec or not though I'm not sure. Can't use your check as it will cause recursion in the fix.

es5-shim.js
```diff
@@ -329,6 +360,97 @@ var strSplit = call.bind(StringPrototype.split);
...
+var hasToStringTagArrayBug = toStr([]) !== '[object Array]';
```
ljharb: in what browser is this `true`? The defacto `Array.isArray` shim, which works in every single browser I've ever tested in, is `toStr(foo) === '[object Array]'`

es5-shim.js
```diff
@@ -329,6 +360,97 @@ var strSplit = call.bind(StringPrototype.split);
...
+var hasToStringTagRegExpBug = toStr(/test/) !== '[object RegExp]';
```
ljharb: similarly, in what browser is this `true`?

es5-shim.js
```diff
@@ -329,6 +360,97 @@ var strSplit = call.bind(StringPrototype.split);
...
+        typeof value.length === 'number' &&
+        value.length >= 0 &&
+        !owns(value, 'callee') &&
+        !owns(value, 'arguments');
+  };
+}
+
+if (hasToStringTagRegExpBug) {
+  var isDuckTypeRegExp = function (value) {
+    /*jshint eqnull:true */
```
ljharb: remove jshint directive

es5-shim.js
```diff
@@ -329,6 +360,97 @@ var strSplit = call.bind(StringPrototype.split);
...
+    if (value != null) {
+      var type = typeof value;
+      if (type === 'object' || type === 'function') {
+        try {
+          regexExec.call(value);
```
ljharb: this should be robust against `delete Function.prototype.call`

es5-shim.js
Xotic750: IE6-8 -> hasToStringTagBasicBug
The others I can't remember exactly, but could pick the tests across and run them as I have no way of testing any more.

es5-shim.js
Xotic750: Not sure what you mean?

ljharb: Any `.call` will fail if `Function.prototype.call` is later broken or nonexistent

Xotic750: Ok.

es5-shim.js
```diff
@@ -163,7 +164,6 @@ var ES = {
     // http://es5.github.com/#x9.9
     /* replaceable with https://npmjs.com/package/es-abstract ES5.ToObject */
     ToObject: function (o) {
-        /* jshint eqnull: true */
```
ljharb: hm, i didn't realize this was in master. thanks for removing it!

es5-shim.js
ljharb: Primarily my concern is keeping the chunks of code that are replaceable with existing npm modules together and untouched.

es5-shim.js
ljharb: yes - please don't add any of the checks unless you can include a comment indicating in which browsers it fails.

ljharb: * specifically, the runtime checks. adding actual spec tests is always OK :-)

es5-shim.js
Xotic750: No problem.

es5-shim.js
Xotic750: Not really sure how this can be achieved. If someone deletes `Function.prototype.call` it seems everything is pretty much fubar, whether es5-shim is loaded or not. Did you have something in mind?

es5-shim.js
```diff
@@ -329,6 +351,94 @@ var strSplit = call.bind(StringPrototype.split);
 var strIndexOf = call.bind(StringPrototype.indexOf);
 var push = call.bind(array_push);
 
+// ES-5 15.3.4.4
+// http://es5.github.io/#x15.3.4.4
+var hasToStringTagBasicBug = toStr() !== '[object Undefined]' || toStr(null) !== '[object Null]';
+var hasToStringTagLegacyArguments = toStr(function () {
+  return arguments;
+}) !== '[object Arguments]';
```
michaelficarra: You forgot to invoke the function.

Xotic750: Dope!

es5-shim.js
```diff
@@ -329,6 +351,94 @@ var strSplit = call.bind(StringPrototype.split);
...
+          return false;
+        }
+      }
+    }
+  };
+}
+
+if (hasToStringTagInconsistency) {
+  defineProperties(FunctionPrototype, {
+    call: function () {
```
michaelficarra: You'll have to do the same thing with `apply`, right? Might want to just have `call` defer to `apply` in that case.

es5-shim.js
ljharb: It's not possible in the sense that the `bind` polyfill relies on it (and has to). However, none of the other functions need to.

In the es5-shim, `.call` is one of the ones I haven't tried too hard to be robust against, but it's still useful to try.

es5-shim.js
Xotic750: Yeah, a similar thing will need to be done with `apply`. Just concentrating on `call` right now, exploring the possibilities, proving that it can be done.

es5-shim.js
```diff
@@ -329,6 +351,94 @@ var strSplit = call.bind(StringPrototype.split);
 var strIndexOf = call.bind(StringPrototype.indexOf);
 var push = call.bind(array_push);
 
+// ES-5 15.3.4.4
+// http://es5.github.io/#x15.3.4.4
+var hasToStringTagBasicBug = toStr() !== '[object Undefined]' || toStr(null) !== '[object Null]';
+var hasToStringTagLegacyArguments = toStr(function () {
+  return arguments;
+}()) !== '[object Arguments]';
+var hasToStringTagArrayBug = toStr([]) !== '[object Array]';
```
ljharb: just making sure we don't forget to delete this line, unless we can find (and document in a comment) a specific browser that would give `true` here.

es5-shim.js
```diff
@@ -329,6 +351,94 @@ var strSplit = call.bind(StringPrototype.split);
...
+var hasToStringTagRegExpBug = toStr(/test/) !== '[object RegExp]';
```
ljharb: just making sure we don't forget to delete this line, unless we can find (and document in a comment) a specific browser that would give `true` here.

es5-shim.js
```diff
@@ -329,6 +351,94 @@ var strSplit = call.bind(StringPrototype.split);
 var strIndexOf = call.bind(StringPrototype.indexOf);
 var push = call.bind(array_push);
 
+// ES-5 15.3.4.4
+// http://es5.github.io/#x15.3.4.4
+var hasToStringTagBasicBug = toStr() !== '[object Undefined]' || toStr(null) !== '[object Null]';
+var hasToStringTagLegacyArguments = toStr(function () {
+  return arguments;
+}()) !== '[object Arguments]';
```
ljharb: in this case, it's already being invoked in an IIFE, so you can just do `toStr(arguments) !== '[object Arguments]'`

es5-shim.js
```diff
@@ -1452,7 +1562,7 @@ defineProperties(NumberPrototype, { toFixed: toFixedShim }, hasToFixedBugs);
 
 var hasToPrecisionUndefinedBug = (function () {
     try {
-        return 1.0.toPrecision(undefined) === '1';
+        return (1.0).toPrecision(undefined) === '1';
```
ljharb: these parens should not be needed. please revert.

es5-shim.js
Xotic750: Good point! :)

es5-shim.js
```diff
@@ -1654,7 +1764,7 @@ defineProperties(StringPrototype, {
     }
 }, hasTrimWhitespaceBug);
 
-var hasLastIndexBug = StringPrototype.lastIndexOf && 'abcあい'.lastIndexOf('あい', 2) !== -1;
+var hasLastIndexBug = StringPrototype.lastIndexOf && 'abc\u3042\u3044'.lastIndexOf('\u3042\u3044', 2) !== -1;
```
ljharb: please revert this change as well - the file is already UTF-8.

tests/spec/s-function.js
```diff
@@ -1,7 +1,177 @@
 /* global describe, it, expect, beforeEach */
+var hasStrictMode = (function () {
+  'use strict';
+
+   return !this;
+}());
+
+var global = (function () {
+   return this;
```
ljharb: this will fail silently if strict mode has been globally enabled - please ensure that `this` is truthy as well.

tests/spec/s-function.js
```diff
@@ -1,7 +1,177 @@
...
+        it('should not throw a error in each case', function () {
+            expect(function () {
+                noop.call();
+            }).not.toThrow();
+            expect(function () {
+                /*eslint no-useless-call: 0*/
+                noop.call(undefined);
+            }).not.toThrow();
+            expect(function () {
+                /*eslint no-useless-call: 0*/
```
ljharb: nbd, but i find it easier to grep for `/* eslint-disable no-useless-call */` and `/* eslint-enable no-useless-call */`

tests/spec/s-function.js
```diff
@@ -1,7 +1,177 @@
...
+            }).not.toThrow();
+            expect(function () {
+                noop.call(true);
+            }).not.toThrow();
+            expect(function () {
+                noop.call(noop);
+            }).not.toThrow();
+            expect(function () {
+                noop.call(new Date());
+            }).not.toThrow();
```
ljharb: i think all these 3-line assertions can be condensed to one line.

es5-shim.js
Xotic750: Ok

tests/spec/s-function.js
```diff
@@ -1,7 +1,177 @@
...
+            expect(function () {
+                noop.call(/test/);
+            }).not.toThrow();
+        });
+        it('should not throw a error with additional args', function () {
```
ljharb: what is this protecting against? all functions in JS accept additional arguments - and this isn't testing that `call` is passing the arguments through

es5-shim.js
Xotic750: Ok

tests/spec/s-function.js
```diff
@@ -1,7 +1,177 @@
...
+                noop.call(true, undefined, null, '1', 1, true, noop);
+            }).not.toThrow();
+            expect(function () {
+                noop.call(noop, undefined, null, '1', 1, true, noop);
+            }).not.toThrow();
+            expect(function () {
+                noop.call(new Date(), undefined, null, '1', 1, true, noop);
+            }).not.toThrow();
+        });
+        if (hasStrictMode) {
```
ljharb: instead of this if-wrapping, please use the approach I've used in most other test files - something like:
```js
var ifHasStrictIt = hasStrictMode ? it : xit;`
```

(I'll fix the current instance of this pattern in the array tests)

tests/spec/s-object.js
```diff
@@ -356,4 +361,35 @@ describe('Object', function () {
...
+            expect(toStr.call([1, 2, 3])).toBe('[object Array]');
+            expect(toStr.call(returnArgs())).toBe('[object Arguments]');
+            expect(toStr.call({})).toBe('[object Object]');
+            expect(toStr.call(noop)).toBe('[object Function]');
+            expect(toStr.call(new RegExp('c'))).toBe('[object RegExp]');
+            expect(toStr.call(new Date())).toBe('[object Date]');
+            expect(toStr.call(new Error('x'))).toBe('[object Error]');
+        });
+
+        it('prototypes', function () {
```
ljharb: this test will actually cause problems in future implementations, since in ES6, many of these prototype objects stop being instances of themselves, and instead would return `'[object Object]'`.

Please split these up such that array/boolean/object/function are always tested, but that the rest are skipped with an `xit`, much like https://github.com/es-shims/es5-shim/blob/master/tests/spec/s-object.js#L89-L90

tests/spec/s-object.js
```diff
@@ -22,6 +22,11 @@ var canFreeze = typeof Object.freeze === 'function' && (function () {
     return obj.a !== 3;
 }());
 var ifCanFreezeIt = canFreeze ? it : xit;
+var toStr = Object.prototype.toString;
+var noop = function () {};
+var returnArgs = function () {
```
ljharb: this can probably just be inlined at the one place it's used?

tests/spec/s-function.js
Xotic750: Aha, I looked at the docs and wasn't sure how to do it, I've never use `eslint` before.

tests/spec/s-function.js
Xotic750: Ok

tests/spec/s-function.js
Xotic750: I threw in all the tests that I had. There are sure to be improvements. I'll try and improve them some.

tests/spec/s-function.js
Xotic750: I'll take a closer look and try to mimic your style.

tests/spec/s-object.js
Xotic750: I'll take a look and try to do that.

tests/spec/s-object.js
Xotic750: Ok

es5-shim.js
Xotic750: I'll change it to use `bind`, at present I can't think of another way.

ljharb:
Thanks for bearing with all the comments! Please rebase this on top of latest master as you make your changes as well.

Xotic750:
Will do. Comments are good, I'm learning. ;)

es5-shim.js
Xotic750: Ok, altered some settings in my editor and changed them back, now I'm seeing a warning `this character may get silently deleted by one or more browsers`. Just thought I'd mention it before I go and disable this warning.

ljharb: Thanks - It's only an issue if text encoding isn't properly set up - I'm not concerned about it at this point.

tests/spec/s-function.js
Xotic750: Like this?
```
var global = (function () {
    if (this) {
        return this;
    }
}());
```

tests/spec/s-function.js
```diff
@@ -1,7 +1,177 @@
...
+                }).not.toThrow();
+                expect(result).toBe(subject);
+            });
+        }
+        it('should have correct context in non-strict mode', function () {
+            var result,
+                testFn,
+                subject;
+
+            testFn = function () {
```
ljharb: we don't use `onevar`, so please use one `var` per var

tests/spec/s-function.js
```diff
@@ -1,7 +1,177 @@
...
+                return this;
+            };
+
+            expect(function () {
+                result = testFn.call();
+            }).not.toThrow();
+
+            expect(result).toBe(global);
```
ljharb: in these tests, why not just `expect(testFn.call()).toBe(global)`? in what scenario would `call` ever throw on a non-throwing function?

tests/spec/s-function.js
ljharb: I'm thinking more that this needs to be a test - since detecting the global object is very difficult.

Alternatively, you could use `var global = Function('return this')();` which is unassailably correct everywhere.

Xotic750: Ok, I'll use your suggestion but I have a feeling I read somewhere that this can call problems in some versions of Chrome, but I'm not sure.

tests/spec/s-function.js
Xotic750: Ok

tests/spec/s-function.js
Xotic750: I'll change it. Again, just all my old naive tests, they've never been scrutinised by anyone before.

tests/spec/s-function.js
ljharb: I used it in the es6-shim for a long time - it can cause problems in Chrome OS plugins, because they disallow CSP, but that's not a concern for tests.

tests/spec/s-function.js
ljharb: in that case the exception would auto-fail the test anyways :-)

Xotic750:
@ljharb Ok, I think I've dealt with all (or at least the majority) of the comments. Feel free to throw some more at me. I still have no idea where I can see the browser tests, so I'm really unable to assess what is going on there but the Node stuff all looks good.

Xotic750:
I managed to get a test to run on IE8 and it failed. I have clearly got something wrong, it doesn't look like the patch is running or something.

es5-shim.js
```diff
@@ -328,6 +351,99 @@ var strSplit = call.bind(StringPrototype.split);
...
+          return '[object Arguments]';
+        }
+        if (hasToStringTagArrayBug && isDuckTypeArray(thisArg)) {
+          return '[object Array]';
+        }
+        if (hasToStringTagRegExpBug && isTryCatchRegExp(thisArg)) {
+          return '[object RegExp]';
+        }
+      }
+      return apply.bind(this, thisArg, baseSlice(arguments, 1))();
```
ljharb: it seems like we should cache `apply.bind(call)` outside the function, and just invoke that directly here?

Xotic750: Probably, I will try to make some changes today and see if I can a slot on IE8, perhaps even force the path with a `true` to see what is happening.

Xotic750:
We now have all green on Opera 11 and IE10.
IE8, tests now run but we have recursive problem due to `bind` shim using `call`. Will think about it a bit.

Xotic750:
IE8 `Function#call` now works and we get correct results with `Object#toString`. I unable to test any other old ES3 browser (IE<9, Opera<11 or any that I have not already mentioned). Will rebase now and have a think about `apply`, looks like a challenge.

Xotic750:
I believe I have a solution for `apply`, but I don't know if it would be acceptable? It would mean extending `Function.prototype` with a new method of some `name` that is unlikely to clash with anything a user may do. e.g `__saAdhVjh5wei3C5fibsd34vGs1udf__` or something like that?

ljharb:
Yeah that's definitely not acceptable :-/ Reflection would easily expose it.

es5-shim.js
```diff
@@ -328,6 +329,73 @@ var strSplit = call.bind(StringPrototype.split);
...
 
+// ES-5 15.3.4.4
+// http://es5.github.io/#x15.3.4.4
+var hasToStringTagBasicBug = toStr() !== '[object Undefined]' || toStr(null) !== '[object Null]';
+var hasToStringTagLegacyArguments = toStr(arguments) !== '[object Arguments]';
+var hasToStringTagInconsistency = hasToStringTagBasicBug || hasToStringTagLegacyArguments;
+
+// Others to check:
+// PhantomJS 1.9 which returns 'function' for `NodeList` instances.
+// Older ES3 native functions like `alert`, there are some fixes available but
```
ljharb: no, i'm not going to worry about older "alert" implementations

es5-shim.js
```diff
@@ -328,6 +329,73 @@ var strSplit = call.bind(StringPrototype.split);
...
+        while (++index < length) {
+            result[index] = array[index + begin];
+        }
+        return result;
+    };
+
+    defineProperties(FunctionPrototype, {
+      call: function () {
+          var thisArg = arguments[0];
+          if (this === to_string) {
```
ljharb: `to_string` doesn't seem to be defined anywhere

es5-shim.js
```diff
@@ -328,6 +329,73 @@ var strSplit = call.bind(StringPrototype.split);
...
+              if (hasToStringTagBasicBug) {
```
ljharb: any reason not to unconditionally enter this code branch? If `call` has to be patched, we might as well just do null and undefined for it

es5-shim.js
Xotic750: https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L63

es5-shim.js
Xotic750: We only want `call` to give us a `[[Class]]` if the context is `Object#toString`, otherwise we want it to go about its normal business of call whatever the function `this` is.

Xotic750:
I'll clean that up, and a variable. I noticed,  that is not used. I'll leave a comment about this is where other native objects can be fixed (even though you don't intend to). I'll also cleanup `baseSlice` into just a loop. I can't see me using it for `apply` for some time, as I can't think of a way to avoid recursion at present without extending `Function#` or attaching (and removing) a property to the individual function so that `apply` can get context.

Xotic750:
The only useful thing remaining, that I can think of, would be for normalising the `[[Class]]` response when checking `window` or `global`. But that is something I can leave you to think(or not) about. ;)

Xotic750:
And I think I've found a solution to the recursion problem and getting context without extending Function# ;)

es5-shim.js
ljharb: ah, my bad, i thought you defined it :-) disregard

ljharb:
Yes, let's not bother with `window` or the global object.

es5-shim.js
```diff
@@ -328,6 +328,54 @@ var strSplit = call.bind(StringPrototype.split);
...
+        !owns(value, 'arguments') &&
+        owns(value, 'callee');
+        // Using `isCallable` here is dangerous as it can cause recursion.
+        // A typeof check would be ok if necessary.
+  };
+}
+
+defineProperties(FunctionPrototype, {
+  call: function () {
+      var thisArg = arguments[0];
```
ljharb: `Function.call` has a length of `1`, so `thisArg` should be a named argument. http://www.ecma-international.org/ecma-262/5.1/#sec-15.3.4.4

es5-shim.js
```diff
@@ -328,6 +328,54 @@ var strSplit = call.bind(StringPrototype.split);
...
+          }
+          if (typeof thisArg === 'undefined') {
+              return '[object Undefined]';
+          }
+          if (hasToStringTagLegacyArguments && isDuckTypeArguments(thisArg)) {
+              return '[object Arguments]';
+          }
+      }
+      var i = arguments.length;
+      var rest = i ? Array(--i) : [];
```
ljharb: please explicitly compare to `0` instead of relying on truthiness - however, i'm not sure why we need `Array()` at all? There's no performance benefit towards "allocating" slots on a JS array, in fact the opposite - because it initially creates a sparse array.

es5-shim.js
Xotic750: Ok

es5-shim.js
Xotic750: I though it was cheaper than using `push`, if you have a large number or arguments?

es5-shim.js
```diff
@@ -56,9 +56,17 @@ var array_push = ArrayPrototype.push;
 var array_unshift = ArrayPrototype.unshift;
 var array_concat = ArrayPrototype.concat;
 var call = FunctionPrototype.call;
+var apply = FunctionPrototype.apply;
 var max = Math.max;
 var min = Math.min;
 
+// To prevent recursion if `Function#call` or `Function#apply` are patched.
```
Xotic750: This actually needs to be moved to happen on when patch is applied. I will fix it an rebase later.

es5-shim.js
```diff
@@ -328,6 +329,70 @@ var strSplit = call.bind(StringPrototype.split);
...
+// Tests for inconsistent or buggy `[[Class]]` strings.
+var hasToStringTagBasicBug = toStr() !== '[object Undefined]' || toStr(null) !== '[object Null]';
+var hasToStringTagLegacyArguments = toStr(arguments) !== '[object Arguments]';
+var hasToStringTagInconsistency = hasToStringTagBasicBug || hasToStringTagLegacyArguments;
+// Others that could be fixed:
+// Older ES3 native functions like `alert` return `[object Object]`.
+// Inconsistent `[[Class]]` strings for `window` or `global`.
+
+if (hasToStringTagLegacyArguments) {
+    // To prevent recursion if `Function#call` or `Function#apply` are patched.
```
Xotic750: This actually seems to have some strange effect on newer environments (when patch is forced) when not used in conjunction with my `apply` patch. I'm thinking that I may remove it from this this PR.

Xotic750:
I am going to remove the `Array#slice` test from this PR and submit those separately.

Xotic750:
I now have apply working on IE8
https://rawgit.com/Xotic750/es5-shim/apply/tests/index.html
Ref: https://github.com/es-shims/es5-shim/issues/99
I feel it is more robust to implement both fixes together. What are your thoughts on how to proceed?

Xotic750:
`call` + `apply` fix is far more stable.

PR #346, opened 2015-11-17
Add tests for `Array#slice`

Xotic750:
Important to be sure of functionality for future `call` and `apply` fixes.

### DIFF
--- a/tests/spec/s-array.js
+++ b/tests/spec/s-array.js
@@ -1540,4 +1540,32 @@ describe('Array', function () {
             expect(obj[2]).toBeUndefined();
         });
     });
+
+    describe('#slice()', function () {
+        it('works on arrays', function () {
+            var arr = [1, 2, 3, 4];
+            var result = arr.slice(1, 3);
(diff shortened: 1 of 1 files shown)


tests/spec/s-array.js
```diff
@@ -1540,4 +1540,30 @@ describe('Array', function () {
...
+            var result = arr.slice(1, 3);
+            expect(result).toEqual([2, 3]);
+        });
+
+        it('is generic', function () {
+            var obj = { 0: 1, 1: 2, 2: 3, 3: 4, length: 4 };
+            var result = Array.prototype.slice.call(obj, 1, 3);
+            expect(result).toEqual([2, 3]);
+
+            obj = (function () {
```
ljharb: let's split this into two "it" tests rather than reusing the variable

ljharb:
Thanks, one comment and I'll pull this in :-)

tests/spec/s-array.js
Xotic750: Ok

PR #347, opened 2015-11-17
Patch `Function#call` and `Function#apply` together, more robust than single fix.

Xotic750:
`call` Ref: https://github.com/es-shims/es5-shim/issues/304
`apply` Ref: https://github.com/es-shims/es5-shim/issues/99
Original `call` only fix: https://github.com/es-shims/es5-shim/pull/345

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -55,6 +55,7 @@ var array_splice = ArrayPrototype.splice;
 var array_push = ArrayPrototype.push;
 var array_unshift = ArrayPrototype.unshift;
 var array_concat = ArrayPrototype.concat;
+var str_split = StringPrototype.split;
 var call = FunctionPrototype.call;
 var apply = FunctionPrototype.apply;
 var max = Math.max;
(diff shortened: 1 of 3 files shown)


Xotic750:
I am guess that you will want so more tests added for `apply`, like those for `call`. To be honest though, if there was a problem then it would be rearing its ugly head.

es5-shim.js
```diff
@@ -175,11 +177,160 @@ var ES = {
...
+        }
+    }
+    return false;
+}());
+
+var shouldPatchCallApply = hasToStringTagInconsistency || hasApplyArrayLikeDeficiency;
+
+if (shouldPatchCallApply) {
+    // Constant. ES3 maximum array length.
+    var MAX_ARRAY_LENGTH = 4294967295;
```
ljharb: how is this calculated? Is it the same in every ES3 browser, or does it vary?

Xotic750: `Math.pow(2, 32) - 1`. This is the maximum value that any ES3 environment that I have tested supports (I don't know if it is in the spec somewhere). Then in ES5 it changes to `Number.MAX_SAFE_INTEGER` for arrays and strings (this is in the spec as far I remember). Arrays seem to follow this but strings still seem to be limited to 65k characters, it is a similar story with arguments, I don't think any environment supports more than 65k args (and many support much less). So this seemed a sensible number to choose to check if it was array-like, the smallest max array length? We could set it to `Number.MAX_SAFE_INTEGER`, but this patch is only going to be applied on ES3 environments, I don't know, I am in two minds about what it should be?

Xotic750:
Since adding the apply tests, there are 4 failures on IE8 that I am not sure about and need to find out why they happen.

es5-shim.js
Xotic750: I have removed this, well the use of it. Need to make sure and remove the code.

Xotic750:
Just 1 IE8 failing test remaining.

Xotic750:
IE8 is now without fail. IE7 also tested without fail. Opera 11 and 12 tested ok. Need to test other ES3 environments that I don't have access to.

es5-shim.js
```diff
@@ -175,11 +176,143 @@ var ES = {
...
+        // If `this` is `Object#toString`, captured or modified.
+        if (this === to_string || this === Object.prototype.toString) {
+            return toStringTag(thisArg);
+        }
+        // All other applys.
+        if (arguments.length > 1 && type === 'object' && argsArray && argsArray.length > 0) {
+            // Boxed string access bug fix.
+            if (splitString && to_string.call(argsArray) === '[object String]') {
+                // `str_split` is safe to use here, no known issue at present.
+                argsArray = call.call(str_split, argsArray, '');
```
ljharb: there's already a call-bound `strSplit` btw, you should just be able to do `strSplit(argsArray, '')`

ljharb: ah wait, i see that it's defined farther down.

I wonder if we should do all the call-bindings before patching `Function#call`?

Xotic750: I don't think i will make any difference, when `call` is patched then `bind` patch is in play and it will use the patched `call`. When `call` is not patched then there is also a native `bind`. Totally up to you and how you feel about it, do it before or later. The important thing is to not use any bound functions within the patch itself otherwise you end up with a circular condition.

Xotic750:
Is there is anything more that you require for this patch (except a rebase)?

ljharb:
In general this one makes me very nervous, so I'm intending to let it sit and marinate for awhile. I'll want to extensively test it across many browsers.

I'm also curious to see if there's a performance impact on the browsers this patches, and what it might be.

Xotic750:
No problem, just say if there are any changes that you want.

ljharb:
@Xotic750 can you please check the "allow edits" checkbox on the RHS of the PR?

PR #349, opened 2015-11-19
Fix `Array#slice` boxed string access on IE<9

Xotic750:


ljharb:
Does that mean the current tests fail on IE 9?

I'll have to review all this stuff this weekend; this is a busy week for me.

Xotic750:
On IE8 and IE7 (IE<9), `Array#slice` failed on string literal test that was added. IE9 passed ok.

ljharb:
Thanks, fix merged!

PR #351, opened 2015-11-19
PropertyIsEnumerable and hasOwnProperty method fails when you use the Object.create function in Internet Explorer

Xotic750:
https://support.microsoft.com/en-gb/kb/2980015
http://webreflection.blogspot.se/2014/04/all-ie-objects-are-broken.html
IE9 on saucelabs failed, but IE 10 & 11 have been patched and passed.

### DIFF
--- a/tests/spec/s-object.js
+++ b/tests/spec/s-object.js
@@ -355,5 +355,12 @@ describe('Object', function () {
 
             expect(obj instanceof Object).toBe(false);
         });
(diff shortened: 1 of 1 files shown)


ljharb:
The "update 3" on your first link indicates that IE 9 is indeed patched, and your test passes for me on IE 9 on Browserstack.

In light of this, and because the patch for it isn't bulletproof, I'm going to close this for now.

PR #353, opened 2015-11-19
Add tests for `Array#reverse`, IE<9 fails some of these

Xotic750:


tests/spec/s-array.js
```diff
@@ -1568,4 +1568,141 @@ describe('Array', function () {
...
+                true,
+                false,
+                1,
+                0,
+                -1,
+                null,
+                undefined
+            ));
+            Array.prototype.reverse.call(arrCmp);
+            expect(Object.prototype.toString.call(arrCmp)).toBe('[object Arguments]');
```
ljharb: this test can't ever pass in IE < 9, because it has legacy arguments objects, and we don't patch `Object#toString` (nor could we, reliably)

tests/spec/s-array.js
```diff
@@ -1568,4 +1568,141 @@ describe('Array', function () {
...
+                false,
+                1,
+                0,
+                -1,
+                null,
+                undefined
+            ];
+            Array.prototype.reverse.call(arrCmp);
+            expect(Object.prototype.toString.call(arrCmp)).toBe('[object Object]');
+            expect(arrCmp).toEqual(expected);
```
ljharb: this test only fails because you started with an object, and `expected` isn't an object - `reverse` isn't supposed to magically turn something into an array, that's not what "generic" means.

ljharb:
Based on my tests, no engine fails any of these tests except IE 6 - 8, and they only legitimately fail on the one test that can't ever pass due to a legacy arguments object. I'm going to close this PR for now, unless you know of other engines something fails on?

tests/spec/s-array.js
Xotic750: With the `call` fix this would work reliably (well, as reliable as duck-typing is), but I can change this to an `isArguments` function.

tests/spec/s-array.js
Xotic750: No, expected should have been an object and not an array. Think that was a copy paste error on my part.

Xotic750:
Ok, I'll make some changes here, time allowing, so I can run a few tests again. If I find something then I'll let you know.

ljharb:
Thanks!

PR #354, opened 2015-11-19
Add tests for `Array#sort`

Xotic750:
Throws on all versions of IE, and some other environments.

### DIFF
--- a/tests/spec/s-array.js
+++ b/tests/spec/s-array.js
@@ -1596,4 +1596,376 @@ describe('Array', function () {
             expect(result).toEqual(['2', '3']);
         });
     });
+
+    describe('#sort()', function () {
+        describe('usage', function () {
+            it('requires a function or undefined as first argument', function () {
+                var actual = [];
+                expect(actual.sort()).toBe(actual);
+                expect(actual.sort(undefined)).toBe(actual);
+                expect(actual.sort(function () { return 0; })).toBe(actual);
+            });
+
+            it('requires a non-function or non-undefined to throw a `TypeError`', function () {
+                expect(function () { [].sort(null); }).toThrow();
+                expect(function () { [].sort(1); }).toThrow();
+                expect(function () { [].sort(''); }).toThrow();
+                expect(function () { [].sort(true); }).toThrow();
+                expect(function () { [].sort({}); }).toThrow();
+                expect(function () { [].sort([]); }).toThrow();
+                expect(function () { [].sort(new Date()); }).toThrow();
+                expect(function () { [].sort(/pattern/); }).toThrow();
+            });
+        });
+
+        describe('ascending', function () {
+            it('[5,2,4,6,1,3] should result in [1,2,3,4,5,6]', function () {
+                var actual = [5, 2, 4, 6, 1, 3];
+                var expected = [1, 2, 3, 4, 5, 6];
+                actual.sort();
+                expect(actual).toEqual(expected);
+            });
+
+            it('[5,2,2,6,1,3] should result in [1,2,2,3,5,6]', function () {
+                var actual = [5, 2, 2, 6, 1, 3];
+                var expected = [1, 2, 2, 3, 5, 6];
+                actual.sort();
+                expect(actual).toEqual(expected);
+            });
+
+            it('[0,0,0,0,0,1] should result in [0,0,0,0,0,1]', function () {
+                var actual = [0, 0, 0, 0, 0, 1];
+                var expected = [0, 0, 0, 0, 0, 1];
+                actual.sort();
+                expect(actual).toEqual(expected);
+            });
+
+            it('[0,0,0,0,0,-1] should result in [-1,0,0,0,0,0]', function () {
+                var actual = [0, 0, 0, 0, 0, -1];
+                var expected = [-1, 0, 0, 0, 0, 0];
+                actual.sort();
+                expect(actual).toEqual(expected);
+            });
+
+            it('[f,e,d,a,c,b] should result in [a,b,c,d,e,f]', function () {
+                var actual = ['f', 'e', 'd', 'a', 'c', 'b'];
+                var expected = ['a', 'b', 'c', 'd', 'e', 'f'];
+                actual.sort();
+                expect(actual).toEqual(expected);
+            });
+
+            it('[f,e,d,,,,a,c,b] should result in [a,b,c,d,e,f,,,]', function () {
+                var actual = ['f', 'e', 'd', 1, 2, 'a', 'c', 'b'];
+                delete actual[3];
+                delete actual[4];
+                var expected = ['a', 'b', 'c', 'd', 'e', 'f'];
+                expected.length = 8;
+                actual.sort();
+                expect(actual).toEqual(expected);
+            });
+
+            it('[f,e,d,,null,,a,c,b] should result in [a,b,c,d,e,f,null,,,]', function () {
+                var actual = ['f', 'e', 'd', 1, null, 2, 'a', 'c', 'b'];
+                delete actual[3];
+                delete actual[5];
+                var expected = ['a', 'b', 'c', 'd', 'e', 'f', null];
+                expected.length = 9;
+                actual.sort();
+                expect(actual).toEqual(expected);
+            });
+
+            it('[f,e,d,,null,undefined,a,c,b] should result in [a,b,c,d,e,f,null,undefined,,]', function () {
+                var actual = ['f', 'e', 'd', 1, null, undefined, 'a', 'c', 'b'];
+                delete actual[3];
+                var expected = ['a', 'b', 'c', 'd', 'e', 'f', null, undefined];
+                expected.length = 9;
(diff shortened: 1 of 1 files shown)


Xotic750:
I have a drop in replacement already coded that fixes these differences, plus it is a stable sort algorithm. But it's probably too much for es5-shim as stability is not part of the spec.

ljharb:
These tests pass in IE 8, but one of them fails in IE 9-11 (the `[].sort(null)` one), and one fails in IE 6 and 7 ("works with arguments") - but this is an invalid test in that it's impossible to fix legacy arguments objects' `Object#toString`.

tests/spec/s-array.js
```diff
@@ -1568,4 +1568,301 @@ describe('Array', function () {
...
+            });
+        });
+        it('works with `arguments`', function () {
+            var args = function () {
+                return arguments;
+            };
+            var actual = args(5, 2, 4, 6, 1, 3);
+            var expected = [1, 2, 3, 4, 5, 6];
+            Array.prototype.sort.call(actual);
+            expect(Object.prototype.toString.call(actual)).toBe('[object Arguments]');
```
ljharb: these checks aren't relevant to `sort` itself, and can't ever pass in IE 6 - 8.

tests/spec/s-array.js
```diff
@@ -1568,4 +1568,301 @@ describe('Array', function () {
...
+            } else if (leftS < rightS) {
+                val = 1;
+            } else {
+                val = -1;
+            }
+            return val;
+        };
+        var testArr1 = [];
+        for (var i = 0; i < 100; i += 1) {
+            testArr1.push(('00' + (Math.random() * 100).toFixed(0)).slice(-3));
```
ljharb: please use `beforeEach` for setup like this - also it'd be preferable to confine that to a "describe" for only the tests that need it.

tests/spec/s-array.js
```diff
@@ -1568,4 +1568,301 @@ describe('Array', function () {
...
+        var testArr2 = testArr1.slice();
+        it('requires a function or undefined', function () {
+            [].sort();
+            [].sort(undefined);
+            [].sort(null);
+            [].sort(1);
+            [].sort('');
+            [].sort(true);
+        });
+        it('sorting [5,2,4,6,1,3] ascending should result in [1,2,3,4,5,6]', function () {
```
ljharb: please add a line break between every "it"

tests/spec/s-array.js
```diff
@@ -1568,4 +1568,301 @@ describe('Array', function () {
...
+            }
+            return val;
+        };
+        var testArr1 = [];
+        for (var i = 0; i < 100; i += 1) {
+            testArr1.push(('00' + (Math.random() * 100).toFixed(0)).slice(-3));
+        }
+        var testArr2 = testArr1.slice();
+        it('requires a function or undefined', function () {
+            [].sort();
```
ljharb: please make assertions about these lines, rather than just relying on throwing or not. Specifically, that they should return the original array.

tests/spec/s-array.js
```diff
@@ -1568,4 +1568,301 @@ describe('Array', function () {
...
+            var actual = {
+                0: 5,
+                1: 2,
+                2: 4,
+                3: 6,
+                4: 1,
+                5: 3
+            };
+            Array.prototype.sort.call(actual);
+            expect(Object.prototype.toString.call(actual)).toBe('[object Object]');
```
ljharb: what is this testing? if you want to ensure it's not an array, you could use `Array.isArray`, or, assert that `.not.toBe('[object Array]')`?

tests/spec/s-array.js
```diff
@@ -1568,4 +1568,301 @@ describe('Array', function () {
...
+            expect(actual).toEqual({
+                0: 5,
+                1: 2,
+                2: 4,
+                3: 6,
+                4: 1,
+                5: 3
+            });
+
+            actual = {
```
ljharb: this seems like a second test, which means it should have its own "it"

Xotic750:
I will go through this, as soon as I have a spare moment, and clean things up as per comments. ;)

tests/spec/s-array.js
```diff
@@ -1596,4 +1596,385 @@ describe('Array', function () {
...
+            it('requires a function or undefined as first argument', function () {
+                var actual = [];
+                expect(actual.sort()).toBe(actual);
+                expect(actual.sort(undefined)).toBe(actual);
+
+                // http://ecma-international.org/ecma-262/5.1/#sec-15.4.4.11
+                // If comparefn is not undefined and is not a consistent comparison
+                // function for the elements of this array (see below), the behaviour
+                // of sort is implementation-defined.
+                // Some consitency would good here. Throw or not to throw.
```
ljharb: consistency

tests/spec/s-array.js
```diff
@@ -1596,4 +1596,385 @@ describe('Array', function () {
...
+                expect(function () {
+                    actual.sort(null);
+                }).not.toThrow();
```
ljharb: instead of asserting these shouldn't throw, can instead they be `expect(actual.sort(null)).toBe(actual)` like the undefined tests above?

tests/spec/s-array.js
Xotic750: I will remove it, it was kind of a note to self.

tests/spec/s-array.js
Xotic750: Yeah, I guess so.

ljharb:
This branch has a merge commit - could you instead rebase it on top of master?

Xotic750:
I'm no sure how to do that? Everything else I have tried (other than what I'm doing) ends up with all the commits to master in the PR along with my changes.

ljharb:
`git checkout sort && git fetch upstream && git rebase upstream/master && git push -u origin sort -f` should do it.

Xotic750:
Ok, I'll give that a try.

Xotic750:
Cool, something else that have learnt today. :)

ljharb:
OK so - reading http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.11. It does say "If comparefn is not undefined and is not a consistent comparison function for the elements of this array (see below), the behaviour of sort is implementation-defined." but also, step 13.a. says it should throw for any non-function non-undefined argument.

Chrome and Safari (and IE < 9) seem to ignore the argument, however, Firefox and IE >= 9 seem to throw.

I'm not convinced that this is something that should be patched - but, if it was patched, I think it should be patched to always throw when a non-undefined non-function is passed. I'm confirming with the spec folks.

Xotic750:
To throw was my feeling on he subject.

ljharb:
This indeed appears to be an ES5 violation - it should throw. See https://github.com/tc39/ecma262/issues/241 for details.

Can you modify the tests to assert that it does throw?

Xotic750:
Ok, I'll change those back to expect.toThrow

tests/spec/s-array.js
```diff
@@ -1596,4 +1596,439 @@ describe('Array', function () {
...
+                        expect(e).toEqual(jasmine.any(TypeError));
+                        throw e;
+                    }
+                }).toThrow();
+
+                expect(function () {
+                    try {
+                        [].sort({});
+                    } catch (e) {
+                        expect(e).toEqual(jasmine.any(TypeError));
```
ljharb: instead of explicitly doing this assertion, you can do `.toThrow(TypeError)` and remove a lot of this boilerplate - ideally each of these assertions is on a single line.

Xotic750: I tried that but it didn't appear to work in Jasmine 1.3.1, though it worked fine in Jasmine 2?

Xotic750: I've changed them, maybe I did something wrong when I first tried.

Xotic750: Still not working for me, but I'll leave them like that for now.

Xotic750: Tried `toThrow(jasmine.any(TypeError))` and that also doesn't work for me.

tests/spec/s-array.js
```diff
@@ -1596,4 +1596,376 @@ describe('Array', function () {
...
+        describe('usage', function () {
+            it('requires a function or undefined as first argument', function () {
+                var actual = [];
+                expect(actual.sort()).toBe(actual);
+                expect(actual.sort(undefined)).toBe(actual);
+                expect(actual.sort(function () { return 0; })).toBe(actual);
+            });
+
+            it('requires a non-function or non-undefined to throw a `TypeError`', function () {
+                expect(function () { [].sort(null); }).toThrow(jasmine.any(TypeError));
```
ljharb: hmm, this might require `toThrow(new TypeError())`, and might require a message, but I'm also OK with not specifying in the test what kind of error it throws. just `toThrow` is enough, since no engine that throws throws a different kind of error.

Xotic750: I'll remove the specified error type, that way it'll keeps it small.

PR #357, opened 2015-11-19
Add tests for `hasOwnProperty`

Xotic750:
IE<9 fails due to boxed string access.
There were others I seem to remember Opera 9, I think, but again I no longer have access to these older ES3 environments. But these were all solvable by patches.

### DIFF
--- a/tests/spec/s-object.js
+++ b/tests/spec/s-object.js
@@ -356,4 +356,191 @@ describe('Object', function () {
             expect(obj instanceof Object).toBe(false);
         });
     });
+
+    describe('.hasOwnProperty()', function () {
+        var noop = function () {};
+        var obj = {
+                'toString': noop,
+                'toLocaleString': noop,
+                'valueOf': noop,
+                'hasOwnProperty': noop,
+                'isPrototypeOf': noop,
+                'propertyIsEnumerable': noop,
+                'constructor': noop
+            };
+        var obj2 = {};
+        var hasOwn = Object.prototype.hasOwnProperty;
+
+        it('defined on object "toString"', function () {
+            expect(hasOwn.call(obj, 'toString')).toBe(true);
+        });
+
+        it('defined on object "toLocaleString"', function () {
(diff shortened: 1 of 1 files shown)


tests/spec/s-object.js
```diff
@@ -356,4 +356,191 @@ describe('Object', function () {
...
+
+        it('defined on object with "undefined" value "constructor"', function () {
+            expect({
+                constructor: undefined
+            }.hasOwnProperty('constructor')).toBe(true);
+        });
+
+        it('string defined', function () {
+            var str = 'abc';
+            expect(str.hasOwnProperty('0')).toBe(true);
```
ljharb: I'm pretty sure this isn't fixable via a patch - it's why bracket access on strings fails on IE <= 8, and you need to use `charAt` instead.

tests/spec/s-object.js
```diff
@@ -356,4 +356,191 @@ describe('Object', function () {
...
+            expect(str.hasOwnProperty('1')).toBe(true);
+            expect(str.hasOwnProperty('2')).toBe(true);
+        });
+
+        it('string not-defined', function () {
+            var str = 'abc';
+            expect(str.hasOwnProperty('3')).toBe(false);
+        });
+
+        it('string object defined', function () {
```
ljharb: similarly, patching this would require replacing `Object`, and would give a false report if someone was using this to determine if bracket access on strings is safe

ljharb:
If the only ones of these that fail anywhere we can verify are the string ones in IE 6 - 8, I don't think these tests will add much value. Do you know of any other engines where these fail?

tests/spec/s-object.js
Xotic750: I was pretty sure that I had some successful fixes kicking around somewhere (in rough draft), I'll see if I can dig them up.

tests/spec/s-object.js
Xotic750: Hmmm, that could be a problem and is probably the biggest reason to not fix it.

Xotic750:
No, just IE that I know of.

PR #358, opened 2015-11-19
Add tests to `error` to confirm non-enumerability.

Xotic750:
IE < 9 (IE 8 possibly fixable)
IE9
Opera 11 & 12
FireFox 4-27

Some of this is inconsistencies rather than spec.

### DIFF
--- a/tests/spec/s-error.js
+++ b/tests/spec/s-error.js
@@ -19,10 +19,36 @@ describe('Error', function () {
                 throw new RangeError(msg);
             } catch (e) {
                 error = e;
-            };
+            }
             expect(error.name).toBe('RangeError');
             expect(error.message).toBe(msg);
             expect(String(error)).toBe(error.name + ': ' + msg);
+        });
+    });
+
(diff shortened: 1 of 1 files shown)


tests/spec/s-error.js
```diff
@@ -19,10 +19,36 @@ describe('Error', function () {
...
                 error = e;
-            };
+            }
             expect(error.name).toBe('RangeError');
             expect(error.message).toBe(msg);
             expect(String(error)).toBe(error.name + ': ' + msg);
         });
     });
+
+    describe('should not list ', function () {
```
ljharb: describe and it descriptions should compose like english sentences, but also stand alone. what is this describing? `it message` isn't a sentence. etc.

Xotic750: I will improve it.

ljharb:
Thanks - I'm not touching "stack" anything, and fixing properties on Error instances is impractical, but I can definitely fix the `Error.prototype` ones.

PR #359, opened 2015-11-20
Add a rule to `.gitignore` to ignore Komodo Edit settings for the project.

Xotic750:
This would be very useful so that users using `Komodo Edit` don't accidentally upload their editor settings when working on `es5-shim`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .DS_Store
+*.komodoproject


ljharb:
I don't want to add editor-specific things to gitignore. You can add this to your global gitignore and it will apply to all git projects on your machine.

Xotic750:
Ok, I didn't know that. Back to the docs for a read I guess.

Xotic750:
Thanks for pointing me in the correct direction.
Created `~/.gitignore` and then executed
```
git config --global core.excludesfile '~/.gitignore'
```
Problem solved. ;)

Issue #240, opened 2014-05-19
Date#getUTCDate/getDate/toISOString issues in Opera 10.6


ljharb:
``` js
var d = new Date(-3509827334573292);
assert(d.getMonth() === 0);
assert(d.getUTCMonth() === 0);
```

Both of these fail in Opera 10.6.

ljharb:
Added tests in f05e4236eec7fdefef5f82e731e544acdfed7919

Xotic750:
Just been having a look at this. That's really quite a broken mess. Fixable by replacing all affected methods by the looks of things (quite a few, the 2 you mention, all the `toString` methods, and the milliseconds), but a real mess. Didn't they think to test `Date` before shipping?!!

Xotic750:
I've downloaded and installed Opera 10.6 on my systems and started doing just that. I commented out all the tests, it crashed. I commented out all the code, it ran. I started putting code back in and this line (may not be the only one) caused it to crash when reinstated.
```
var hasToDateStringFormatBug = negativeTestDate.toDateString() !== 'Thu Jan 01 -109252';
```

ljharb:
and it's not try/catch-able?

Xotic750:
Haven't tried catch or anything yet, but `toUTCString` also crashes with the original test dates, but not the new date that I posted

Xotic750:
Nope, try...catch does nothing, it just bums out saying core dump

Xotic750:
I'll see if I can use the original test dates for everything except the ones using `toDateString` and `toUTCString`, but I guess there is little that be done about the core dump

Xotic750:
Whoops, just noticed that I posted these comments here rather than in the PR, sorry!

ljharb:
Updated list of Opera 10.6 failures:
 - #getUTCDate()
  - should return the right value for negative dates
   - Expected -32965446 to be 1, Date(Thu Jan 0-32965446 -109252 03:59:59 GMT-0800).
 - #getDate()
  - should return the right value for negative dates
   - Expected -32965446 to be 1.
 - #toISOString()
  - should return correct dates
   - Expected '-109252-08-78T11:59:59.708Z' to be '-109252-01-01T11:59:59.708Z'.

Xotic750:
10.6
![bs_winxp_opera_10 6](https://cloud.githubusercontent.com/assets/216041/12643386/0db35c6a-c5bc-11e5-972d-c7bf32119528.jpg)

PR #360, opened 2015-11-21
Fixes for non-compliant `Date`, multiple browsers.

Xotic750:
https://github.com/es-shims/es5-shim/issues/240

Add extra tests to check string  and other values.

This highlights differences between environments, i.e. strings. It depends on how far you want to take fixes/normalisation?

A version of `eslint` errors with theses changes, but not on any code that has been added.
```
  1405:19   warning  This function has too many statements (45). Maximum allowed is 30   max-statements
  1536:33   warning  This function has too many statements (33). Maximum allowed is 30   max-statements
  1556:13   error    use `string = String(string)` instead                               no-implicit-coercion
  1598:21   warning  Unary operator '++' used                                            no-plusplus
  1720:39   warning  "searchString" is defined but never used                            no-unused-vars
```
```
            var separatorCopy = new RegExp(separator.source, flags + 'g');
            string += ''; // Type-convert
            if (!compliantExecNpcg) {
```

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -1076,6 +1076,202 @@ defineProperties($Object, {
 // ====
 //
 
+var hasNegativeMonthYearBug = new Date(-3509827329600292).getUTCMonth() !== 0;
+var aNegativeTestDate = new Date(-1509842289600292);
+var aPositiveTestDate = new Date(1449662400000);
+var hasToUTCStringFormatBug = aNegativeTestDate.toUTCString() !== 'Mon, 01 Jan -45875 11:59:59 GMT';
+var hasToDateStringFormatBug;
+var hasToStringFormatBug;
+var timeZoneOffset = aNegativeTestDate.getTimezoneOffset();
+if (timeZoneOffset < -720) {
+    hasToDateStringFormatBug = aNegativeTestDate.toDateString() !== 'Tue Jan 02 -45875';
+    hasToStringFormatBug = !(/^Thu Dec 10 2015 \d\d:\d\d:\d\d GMT[-\+]\d\d\d\d(?: |$)/).test(aPositiveTestDate.toString());
+} else {
+    hasToDateStringFormatBug = aNegativeTestDate.toDateString() !== 'Mon Jan 01 -45875';
+    hasToStringFormatBug = !(/^Wed Dec 09 2015 \d\d:\d\d:\d\d GMT[-\+]\d\d\d\d(?: |$)/).test(aPositiveTestDate.toString());
+}
+
+var originalGetFullYear;
+var originalGetMonth;
+var originalGetDate;
+var originalGetUTCFullYear;
+var originalGetUTCMonth;
+var originalGetUTCDate;
+var originalToUTCString;
+var originalToDateString;
+var originalToString;
+var dayName;
+var monthName;
+var daysInMonth;
+if (hasNegativeMonthYearBug || hasToDateStringFormatBug || hasToUTCStringFormatBug || hasToStringFormatBug) {
+    originalGetFullYear = Date.prototype.getFullYear;
+    originalGetMonth = Date.prototype.getMonth;
+    originalGetDate = Date.prototype.getDate;
+    originalGetUTCFullYear = Date.prototype.getUTCFullYear;
+    originalGetUTCMonth = Date.prototype.getUTCMonth;
+    originalGetUTCDate = Date.prototype.getUTCDate;
+    originalToUTCString = Date.prototype.toUTCString;
+    originalToDateString = Date.prototype.toDateString;
+    originalToString = Date.prototype.toString;
+    dayName = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
+    monthName = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+    daysInMonth = function dim(month, year) {
+       return originalGetDate.call(new Date(year, month, 0));
+    };
+}
+
+defineProperties(Date.prototype, {
+    getFullYear: function getFullYear() {
+        if (!this || !(this instanceof Date)) {
+            throw new TypeError('this is not a Date object.');
+        }
+        var year = originalGetFullYear.call(this);
+        if (year < 0 && originalGetMonth.call(this) > 11) {
+            return year + 1;
+        }
+        return year;
+    },
+    getMonth: function getMonth() {
+        if (!this || !(this instanceof Date)) {
+            throw new TypeError('this is not a Date object.');
+        }
+        var year = originalGetFullYear.call(this);
+        var month = originalGetMonth.call(this);
+        if (year < 0 && month > 11) {
+            return 0;
+        }
+        return month;
+    },
+    getDate: function getDate() {
+        if (!this || !(this instanceof Date)) {
+            throw new TypeError('this is not a Date object.');
+        }
+        var year = originalGetFullYear.call(this);
+        var month = originalGetMonth.call(this);
+        var date = originalGetDate.call(this);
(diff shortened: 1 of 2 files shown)


tests/spec/s-date.js
```diff
@@ -227,6 +240,87 @@ describe('Date', function () {
...
+            var date = new Date(-3509827334573292);
+            expect(date.getHours() + Math.floor(date.getTimezoneOffset() / 60)).toBe(10);
+        });
+    });
+
+    describe('#getMinutes()', function () {
+        it('should return the right value for negative dates', function () {
+            var date = new Date(-3509827334573292);
+            var off = date.getTimezoneOffset();
+            var offHours = Math.floor(off / 60);
```
Xotic750: This may need to be Math.trunc

tests/spec/s-date.js
```diff
@@ -227,6 +240,87 @@ describe('Date', function () {
...
+        it('should return the right value for negative dates', function () {
+            // Opera 10.6/11.61/Opera 12 bug
+            expect(new Date(-3509827334573292).getMonth()).toBe(0);
+        });
+    });
+
+    describe('#getHours()', function () {
+        it('should return the right value for negative dates', function () {
+            var date = new Date(-3509827334573292);
+            expect(date.getHours() + Math.floor(date.getTimezoneOffset() / 60)).toBe(10);
```
Xotic750: This may need to be Math.trunc

tests/spec/s-date.js
```diff
@@ -227,6 +240,87 @@ describe('Date', function () {
...
+        });
+    });
+
+    describe('#getMinutes()', function () {
+        it('should return the right value for negative dates', function () {
+            var date = new Date(-3509827334573292);
+            var off = date.getTimezoneOffset();
+            var offHours = Math.floor(off / 60);
+            var offMins = off - offHours * 60;
+            expect(date.getMinutes() + offMins).toBe(37);
```
Xotic750: I think plus Plus?

es5-shim.js
```diff
@@ -1007,6 +1007,44 @@ defineProperties($Object, {
 // ====
 //
 
+var hasNegtiveMonthYearBug = new Date(-3509827334573292).getUTCMonth() === 12;
```
ljharb: `s/Negtive/Negative`

es5-shim.js
```diff
@@ -1007,6 +1007,44 @@ defineProperties($Object, {
...
+if (hasNegtiveMonthYearBug) {
+    var originalGetFullYear = Date.prototype.getFullYear;
+    var originalGetMonth = Date.prototype.getMonth;
+    var originalGetUTCFullYear = Date.prototype.getUTCFullYear;
+    var originalGetUTCMonth = Date.prototype.getUTCMonth;
+}
+defineProperties(Date.prototype, {
```
ljharb: the `defineProperties` call must be inside the `if` if the things it depends on are defined there.

ljharb:
On Opera 11.6, this passes the `getUTCMonth` test but fails `toUTCString` and `toDateString`. On Opera 12.16, this adds a new `toDateString` failure. On Opera 10.6, this appears to crash my testing browser :-/

Xotic750:
I'll take a further look at this as soon as I get a bit of spare time.  I don't have access to Opera 10.6 but I do the other two.

Xotic750:
My access to any Opera environments is very limited, at times non-existent.  I am still trying to get back to looking at this.

Xotic750:
I was able to test Opera 11.6 and it passed, all green, but that's all I have access to at the moment.

Xotic750:
I manage to test Opera 12 and that fails the `toUTCString` test, but no fault of the code, the problem is native it has a ',' in the string. So this could be tested and patched, maybe just a `replace` if you don't want to replcace the method?

Xotic750:
Been trying to get on `browsershots`, they have enable me a year free account but my browsers crash when I try to use it, so can't test Opera 10.6

Xotic750:
Managed to get on Opera 10.6 and it crashed when I ran tests. :/

Xotic750:
This patch now fixes `toDateString`, `toUTCString` and `toString` but it doesn't tackle `toLocal` methods. IE7-9 now give correct strings but need to test more browsers for any issues though. Opera 10.6 still crashes and I have no idea why?

Xotic750:
No matter what I've tried, I am unable to get IE11 and 20 to take the `toUTCString` fix, they continue to use the native method, any ideas?

Xotic750:
Ok, I think I've figured out what is going on with IE11 and 20 now, it concerns their representation of negative dates.

Xotic750:
That now fixes everything that I can test, except as mention the `toLocal` methods, still don't know what to make of the Opera 10.6 crash or what to do about it?

Xotic750:
Any ideas about discovering the problem with Opera 10.6?

ljharb:
None, sadly. All I can think of is commenting parts out until it doesn't crash, and narrowing down which line does it.

Xotic750:
I've downloaded and installed Opera 10.6 on my systems and started doing just that. I commented out all the tests, it crashed. I commented out all the code, it ran. I started putting code back in and this line (may not be the only one) caused it to crash when reinstated.
```
var hasToDateStringFormatBug = negativeTestDate.toDateString() !== 'Thu Jan 01 -109252';
```
I changed the negative date to
```
var negativeTestDate = new Date(-1509842286000292);
```
and uncommented the line and it now runs, will try reinstating the rest of the code. But it looks like it doesn't like a date that far in the past with `toDateString`.

Haven't tried catch or anything yet, but `toUTCString` also crashes with the original test dates, but not the new date that I posted, and in my terminal it just says 
```
[graham@tyr bin]$ ./opera
Bus error (core dumped)
```
But I can't find a core dump file.

Nope, try...catch does nothing, it just bums out saying core dump.

I'll see if I can use the original test dates for everything except the ones using `toDateString` and `toUTCString`, but I guess there is little that can be done about the core dump

Xotic750:
Opera 10.6 now runs without crashing. I added some canned data for months, and I feel it could really use some more for days of the month: just to be 100% sure that all is working, but I don't have time just now. I haven't performed any browser tests on this rebase, other than Opera 10.6

ljharb:
What was the change that prevented/caused the crash?

Xotic750:
I changed the test date for these detections
```
// original test date
var hasNegativeMonthYearBug = new Date(-3509827329600292).getUTCMonth() !== 0;
// these two changed because they crashed with original test date
var hasToDateStringFormatBug = new Date(-1509842289600292).toDateString() !== 'Mon Jan 01 -45875';
var hasToUTCStringFormatBug = new Date(-1509842289600292).toUTCString() !== 'Mon, 01 Jan -45875 11:59:59 GMT';
```

Xotic750:
Added tests for first and last day of each month, not tested browsers.

Xotic750:
I was just running a test or two, Opera 11.64 on Window 7 fails 3 tests
https://saucelabs.com/beta/tests/1236492e6c284a9baa456898a03dcd6d/watch
So I downloaded it for my system (Fedora linux) to investigate, and lo and behold it passes all the tests.
https://youtu.be/zyT12Fc2eSA
What to make of that?

Xotic750:
Added a fix for the weirdness and now passes Opera 11.64 on Windows and Linux.

Xotic750:
That should fix it for the timezone differences, tested on my local browsers only, where I have control over the timezone offset.

Xotic750:
Opera 11.64 on windows now appears broken again `getUTCMonth` (1 fail), while working fine on Linux

Xotic750:
I seems that the Saucelabs Opera 11.64 is either not emptying its cache or their is a proxy that is serving cached files. This appears to be the reason for the difference.

Xotic750:
I believe that this patch is ready to go, unless you can find any issues with it?

ljharb:
All set! There's still some 10.6 issues, but not new ones. I've left those in #240.

PR #361, opened 2015-11-21
Fixed wsRegexChars used twice

softwarespot:


es5-shim.js
```diff
@@ -1639,8 +1639,8 @@ var ws = '\x09\x0A\x0B\x0C\x0D\x20\xA0\u1680\u180E\u2000\u2001\u2002\u2003' +
     '\u2029\uFEFF';
 var zeroWidth = '\u200b';
 var wsRegexChars = '[' + ws + ']';
-var trimBeginRegexp = new RegExp('^' + wsRegexChars + wsRegexChars + '*');
-var trimEndRegexp = new RegExp(wsRegexChars + wsRegexChars + '*$');
+var trimBeginRegexp = new RegExp('^' + wsRegexChars + '*');
```
ljharb: This is not actually functionally equivalent - because `*` means zero or more. If this was a `+` instead, that would be the same as what's on master. This makes the replacement less efficient by doing a replace even when there is no match.

ljharb:
Unless there's a specific bug this fixes or performance metrics this improves, I prefer it as it is.

PR #364, opened 2015-12-09
Tests for RegExp, toString should return correct flags and in correct order

Xotic750:
IE8 fails this.
Can be fixed by creating from `source` and using flags.
```
  function regExpToString(value) {
    var str = '/' + value.source + '/';
    if (value.global) {
      str += 'g';
    }
    if (value.ignoreCase) {
      str += 'i';
    }
    if (value.multiline) {
      str += 'm';
    }
    if (value.sticky) {
      str += 'y';
    }
    return str;
  }
```

### DIFF
--- a/tests/spec/s-regexp.js
+++ b/tests/spec/s-regexp.js
@@ -1,0 +1,26 @@
+/* global describe, it, expect */
+
+describe('RegExp', function () {
+    'use strict';
+
+    describe('#toString()', function () {
+        describe('literals', function () {
+          it('should return correct flags and in correct order', function () {
+              expect(/pattern/.toString()).toBe('/pattern/');
+              expect(/pattern/i.toString()).toBe('/pattern/i');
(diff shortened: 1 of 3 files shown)


tests/spec/s-regexp.js
```diff
@@ -0,0 +1,15 @@
+/* global describe, it, expect */
+
+describe('RegExp', function () {
+    'use strict';
+
+    describe('.toString()', function () {
```
ljharb: `.` is for static methods - this is an instance method, so it should use `#`

ljharb:
Can you also add an `it` with a bunch of `RegExp` objects?

Xotic750:
Ok

ljharb:
it looks like your rebase deleted the code that adds the regexp tests to the HTML files?

Xotic750:
Ah bugger :)

Xotic750:
Fixed

ljharb:
Thanks!

PR #368, opened 2015-12-13
Add Object.getOwnPropertyNames tests

Xotic750:
I'm experiencing some issues since loading `es6-shim` on my projects, I believe the issue is with `Object.getOwnPropertyNames` but I'm not sure if it is an ES5 or an ES6 issue yet, or maybe both. I'm still unable to get my fork building on Travis CI and need to PR it to begin figuring things out. It maybe that this ends up being closed here and opened over in es6-shim.

### DIFF
--- a/tests/spec/s-object.js
+++ b/tests/spec/s-object.js
@@ -1,4 +1,4 @@
-/* global describe, it, xit, expect, beforeEach, jasmine, window */
+/* global describe, xdescribe, it, xit, expect, beforeEach, jasmine, window, caller, ArrayBuffer, Uint8Array, DataView */
 
 var supportsDescriptors = Object.defineProperty && (function () {
     try {
@@ -32,6 +32,28 @@ var canFreeze = typeof Object.freeze === 'function' && (function () {
     return obj.a !== 3;
 }());
 var ifCanFreezeIt = canFreeze ? it : xit;
+
+var supportsOwnPropertyNames = supportsDescriptors;
+var supportsFunctionName = (function test() {}).name === 'test';
+
+var xdescribeGOPN = supportsOwnPropertyNames ? describe : xdescribe;
+var itHasMap = typeof Map === 'function' ? it : xit;
+var itHasSet = typeof Set === 'function' ? it : xit;
+var itHasArrayBuffer = typeof ArrayBuffer === 'function' ? it : xit;
+var itHasUint8Array = typeof Uint8Array === 'function' ? it : xit;
+var itHasDataView = typeof DataView === 'function' ? it : xit;
+var depricatedProps;
+
+var depTest = function () {
+    'use strict';
+
+    depricatedProps = caller && false;
+};
+try {
+    depTest();
+} catch (ignore) {
+    depricatedProps = true;
+}
 
 describe('Object', function () {
     'use strict';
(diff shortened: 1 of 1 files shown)


Xotic750:
Some of the problems show here, but are they fixable, I don't know/not sure yet?

tests/spec/s-object.js
```diff
@@ -365,4 +379,48 @@ describe('Object', function () {
...
+        itHasSet('works on Set', function () {
+          expect(Object.getOwnPropertyNames(new Set([1, 2, 3, 4]))).toEqual([]);
+        });
+
+        itHasArrayBuffer('works on ArrayBuffer', function () {
+          expect(Object.getOwnPropertyNames(new ArrayBuffer(4))).toEqual([]);
+        });
+
+        itHasUint8Array('works on Uint8Array', function () {
+          expect(Object.getOwnPropertyNames(new Uint8Array(4))).toEqual(['0', '1', '2', '3']);
```
ljharb: typed array, arraybuffer, and dataview tests don't belong in es6-shim right now, since it doesn't address either one.

tests/spec/s-object.js
```diff
@@ -33,6 +33,20 @@ var canFreeze = typeof Object.freeze === 'function' && (function () {
 }());
 var ifCanFreezeIt = canFreeze ? it : xit;
 
+// check whether getOwnPropertyNames works if it's given.
+if (supportsDescriptors) {
+    var object = Object.defineProperty({}, 'x', { enumerable: false, value: 1 });
+    var result = Object.getOwnPropertyNames(object, 'x');
+    var getOwnPropertyNamesWorks = result.length === 1 && result[0] === 'x';
```
ljharb: on what engines is this false?

Xotic750: Probably none, it was a 4am thing. :)

tests/spec/s-object.js
Xotic750: Ok, I have a couple of other tests to add to this, but I'm not necessarily expecting any fixes. All of these, in this first push, I am already dealing with in app, but if any fixes come of them then it will be great. One of the problems that shows with es6-shim loaded is `constructor` shows up on Map and Set. I'll add the rest soon. My fork of es6-shim works on Travis without having to PR and I can just link to the results so that you can see.

Xotic750:
This shows the majority of differences that I am seeing with just es5-shim loaded, and then there are others that I see with es6-shim also loaded. Feel free to comment as usual. ;)

tests/spec/s-object.js
ljharb: That'd be a Map and Set test, not a getOwnPropertyNames test :-) thanks for reporting that, I'll see if I can fix it.

ljharb:
Since this is the es5-shim, no code dealing with ES6 constructs, like Typed Arrays, Maps, Sets, DataViews, etc, belongs in this repo.

Please file es6-shim issues (including ES5 features interacting with ES6 features) on es6-shim, and I'll be happy to fix them.

ljharb:
For all these test cases, can you include a comment indicating where it fails? `Object.getOwnPropertyNames` isn't shimmable, so I'd only want to address things that fail in at least one engine.

Xotic750:
I'll split out tests and include some comments over the next week. I tried to exclude any tests where it is shammed, by limiting them to `supportsDescriptors`.

ljharb:
@Xotic750 can you check the "allow edits" box on the RHS of the PR? it'd also be great if you wanted to rebase this so we can get it in :-)

ljharb:
ping @Xotic750

Issue #377, opened 2016-01-25
properlyBoxesContext throws when loaded after prototypejs 1.7.0.0 and older


gkatsev:
See https://jsfiddle.net/gkatsev/oesj5krb/2/ for a live example.
The runtime check ends up invoking prototypejs code and in version 1.7.0.0 and older of the library this ends up throwing an error because a method that prototypejs expects doesn't exist.
version 1.7.1.0 of prototypejs doesn't have this issue because it ends up wrapping the native methods properly in some cases.
es5-shim probably should not throw an error when in such a setting.

Xotic750:
I'm curious as to why you would load prototypejs before es5-shim, rather than vice-versa?

ljharb:
That was my question, but sadly it's on a customer's page that they don't control.

Xotic750:
Perhaps the documents should clearly state that es5-shim (es5-sham, es6-shim and es7.shim etc) are expected to be loaded first and in a particular order as they are patching the JavaScript engine, even though it seems like common sense.

ljharb:
@Xotic750 a PR to add that verbage would be just fine. Regardless, this should be fixed.

gkatsev:
The issue at hand isn't what the correct usage of es5-shim is. It's whether es5-shim should throw an error when used in this way. It probably should not.

ljharb:
Published as [v4.5.1](https://github.com/es-shims/es5-shim/releases/tag/v4.5.1)

gkatsev:
@ljharb thanks.

PR #379, opened 2016-01-26
Add information about patching order when using ES compatibility shim…

Xotic750:
…s, along with an example. https://github.com/es-shims/es5-shim/issues/377

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@
 `es5-shim.js` and `es5-shim.min.js` monkey-patch a JavaScript context to
 contain all EcmaScript 5 methods that can be faithfully emulated with a
 legacy JavaScript engine.
+**Note:** As `es5-shim.js` is designed to patch the native Javascript
+engine, it should be the library that is loaded first.
 
 `es5-sham.js` and `es5-sham.min.js` monkey-patch other ES5 methods as
 closely as possible.  For these methods, as closely as possible to ES5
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -166,6 +168,17 @@ simply `npm install` and `npm test`.
     provisions of this method, which you cannot possibly
     obtain in legacy engines.
 
+### Example of applying ES compatability shims in a browser project
+
+```html
+<script src="es5-shim.js"></script>
+<script src="es5-sham.js"></script>
+<script src="json3.js"></script>
```
ljharb: i don't even think this is necessary, but it should still be done after all the other es shims.

Xotic750: As `json3` is ECMA5.1, then it seems correct to go after `es5-shim`, I believe `es6-shim` also patches the ECMA5.1 JSON object, which won't happen if it were after all the es shims?

ljharb: ah, if json3 is in fact ES5.1 with nothing extra then sure, why not.

we should probably link all of these to cdnjs so that the examples are copypasteable

Xotic750: Well it claims to be, and the tests seem to prove that it is, unlike `JSON2`. I can add the cdn links, `cdnjs` and current versions?

ljharb: cdn links to the "latest" versions seem like the best choice.

Xotic750: There is no `cdnjs` for `es7-shim` yet. And `cdnjs` does not have a "latest" option unfortunately (not that I could find), it's by release version.

README.md
```diff
@@ -166,6 +168,17 @@ simply `npm install` and `npm test`.
...
 
+### Example of applying ES compatability shims in a browser project
+
+```html
+<script src="https://cdnjs.cloudflare.com/ajax/libs/es5-shim/4.5.1/es5-shim.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/es5-shim/4.5.1/es5-sham.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/json3/3.3.2/json3.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/es6-shim/0.34.2/es6-shim.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/es6-shim/0.34.2/es6-sham.min.js"></script>
+<script src="es7-shim.js"></script>
```
ljharb: for this one, you'd want to use https://wzrd.in/standalone/es7-shim@latest which will autoshim

Xotic750: I'll add that.

ljharb:
Thanks!

PR #382, opened 2016-01-27
Use JSON3 from cdnjs instead of JSON2 from lib.

Xotic750:
Use JSON3 instead of JSON2, as JSON3 is ES5 compliant whereas JSON2 is not.


ljharb:
The problem with this approach is that if I update jasmine, I no longer instantly get the updates on the test HTML files. I prefer the way it is now where they're stored in the repo.

Xotic750:
Ok, how about swapping JSON2 for JSON3?

ljharb:
That one i'm fine with, and for using cdnjs

Xotic750:
I need to test this a bit on saucelabs, it may need http rather than https to work on some of their installations.

ljharb:
if so then i'm happy ignoring saucelabs, because "use http" is a bonkers requirement :-)

Xotic750:
There is definitely something going on with IE6 and IE7 when using https rather than http. I assume that they supported https, I'm not sure?

ljharb:
it works just fine for me in IE 6 and IE 7 with https, so i'm not sure what problems you're running into.

Xotic750:
When I use https on my saucelabs account with those browsers, they can't reach the destination. :/
I was thinking maybe change the url from https:// to just // then it should use which ever depending on the original url?

ljharb:
That will break `file://`.

If this is a real problem we'd want to revert this change, and use a local copy of json3.

Xotic750:
I can't get to index.html if I use https, only http. It's not that I can't just get to cdnjs on https. Something weird going on there

ljharb:
probably they have older SSL certs - ie, they're using SSLv3 - and maybe browserstack has worked around that somehow.

Xotic750:
I'd like to use browserstack more, but the connection to their servers is so unreliable for me, they even gave me a year of free account, which is a real shame.

Issue #341, opened 2015-11-08
3 Array#splice tests failing in IE 8


ljharb:
These tests pass just fine in IE 6 and 7.

Relates to #114.

Xotic750:
This was one persons study and conclusions.
https://social.msdn.microsoft.com/Forums/ie/en-US/6e946d03-e09f-4b22-a4dd-cd5e276bf05a/splice-function-in-ie8-is-broken?forum=iewebdevelopment

Also, IE <9 has a bug/difference where the second argument is required rather than optional. I don't think you have any tests for this.

So, it looks like 2 tests and fixes.

IE8 should just use the full replacement, I don't think you can do a partial.

IE<8  requires a partial for the second argument, when it is not supplied.

PR #385, opened 2016-01-28
Add tests for Array#splice when only the first argument is provided.

Xotic750:
https://github.com/es-shims/es5-shim/issues/341

### DIFF
--- a/tests/spec/s-array.js
+++ b/tests/spec/s-array.js
@@ -1441,6 +1441,21 @@ describe('Array', function () {
             expect(test).toEqual(test2);
         });
 
+        it('should work without optional arguments 1', function () {
+            var array = [0, 1, 2];
+            expect(array.splice(0)).toEqual([]);
+        });
+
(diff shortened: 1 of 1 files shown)


Xotic750:
I am actually wondering what the correct results should be as I get different answers from multiple environments.

ljharb:
I believe this is actually a change from ES5 to ES6, which would mean that es5-shim would *not* check or test for deviations from ES6 semantics. I'll leave this open to verify later.

Xotic750:
Ie11 is clearly in the wrong whichever :P
But an interesting mix of results.

ljharb:
In latest Safari and Chrome and Firefox, these tests fail natively, so I think these are invalid tests.

However, when I update your tests so they're not assuming that splice returns the receiver - since it in fact returns the removed elements - they all pass. I'll update this and land.

ljharb:
@Xotic750 oops, i can't; can you please check the "allow edits" box on the RHS of the PR?

PR #394, opened 2016-04-08
Check if property descriptor is configurable before re-defining it

### LINKED ISSUE #373: Kindle Fire HD 8.9: Uncaught TypeError: Cannot redefine property: defineProperty
I use es5-shim as node module and making build via webpack.

For Kindle Fire HD 8.9 Android 4.0.3 I've got this error:
```undefined: Uncaught TypeError: Cannot redefine property: defineProperty```

Here is the context: 
```javascript
	if (supportsDescriptors) {
	    var ensureNonEnumerable = function (obj, prop) {
	        if (isEnum(obj, prop)) {
	            var desc = Object.getOwnPropertyDescriptor(obj, prop);
	            desc.enumerable = false;
	            Object.defineProperty(obj, prop, desc); // <------------ it happened here!
	        }
	    };
	    ensureNonEnumerable(Error.prototype, 'message');
	    if (Error.prototype.message !== '') {
	      Error.prototype.message = '';
```

User-Agent: Mozilla/5.0 (Linux; U; Android 4.0.3; en-us; WFJWAE Build/IML74K) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 Safari/534.30

Could you please fix it?
Thanks :)

jdelStrother:
Android Browser 4.0 appears to have a non-configurable Error.prototype.name property - 

```js
Object.getOwnPropertyDescriptor(Error.prototype, 'name')
// returns - 
{"value":"Error", "writable": false, "enumerable": false, "configurable": false}
```

and will raise a (confusingly-name) "Cannot redefine property: defineProperty" error if you try to re-define Error.prototype.name.

How about checking for `desc.configurable` before attempting to call `Object.defineProperty` ?

This fixes #373

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -2017,8 +2017,10 @@ if (supportsDescriptors) {
     var ensureNonEnumerable = function (obj, prop) {
         if (isEnum(obj, prop)) {
             var desc = Object.getOwnPropertyDescriptor(obj, prop);
-            desc.enumerable = false;
-            Object.defineProperty(obj, prop, desc);
+            if (desc.configurable) {
+              desc.enumerable = false;
+              Object.defineProperty(obj, prop, desc);
+            }
         }
     };
     ensureNonEnumerable(Error.prototype, 'message');


ljharb:
Hmm - while I think this is a good fix, that property descriptor says it's not enumerable. What does `Object.prototype.propertyIsEnumerable.call(Error.prototype, 'name')` say on Android Browser 4.0?

jdelStrother:
Ugh, sorry - I've been limited to debugging this via lots of `alert()`s, and had to manually copy that message across - I must have typoed it.   The real value is : 

```js
Object.getOwnPropertyDescriptor(Error.prototype, 'name')
// returns - 
{"value":"Error", "writable": false, "enumerable": true, "configurable": false}
```

(and FWIW, yes, `Object.prototype.propertyIsEnumerable.call(Error.prototype, 'name')` returns true)

ljharb:
Phew, ok, at least they're consistent :-D That would be far worse. Thanks!

PR #402, opened 2016-06-10
Fix parseInt and parseFloat

ValeryVS:
parsing with undefined, null or NaN should return NaN

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -1980,7 +1980,7 @@
         parseInt = (function (origParseInt) {
             var hexRegex = /^[\-+]?0[xX]/;
             return function parseInt(str, radix) {
-                var string = trim(str);
+                var string = trim(String(str));
                 var defaultedRadix = $Number(radix) || (hexRegex.test(string) ? 16 : 10);
                 return origParseInt(string, defaultedRadix);
             };
@@ -1992,7 +1992,7 @@
         /* global parseFloat: true */
         parseFloat = (function (origParseFloat) {
             return function parseFloat(string) {
-                var inputString = trim(string);
+                var inputString = trim(String(string));
                 var result = origParseFloat(inputString);
                 return result === 0 && strSlice(inputString, 0, 1) === '-' ? -0 : result;
             };
--- a/tests/spec/s-global.js
+++ b/tests/spec/s-global.js
@@ -67,6 +67,12 @@ describe('global methods', function () {
             expect(parseInt('+0xF')).toBe(15);
             expect(parseInt('+0xF', 16)).toBe(15);
         });
+
+        it('NaN parsing', function () {
+            expect(parseInt(undefined)).toBeNaN();
+            expect(parseInt(null)).toBeNaN();
+            expect(parseInt(NaN)).toBeNaN();
+        });
         /* eslint-enable radix */
     });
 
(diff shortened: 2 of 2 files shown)


es5-shim.js
```diff
@@ -1980,9 +1980,15 @@
         parseInt = (function (origParseInt) {
             var hexRegex = /^[\-+]?0[xX]/;
             return function parseInt(str, radix) {
-                var string = trim(str);
-                var defaultedRadix = $Number(radix) || (hexRegex.test(string) ? 16 : 10);
-                return origParseInt(string, defaultedRadix);
```
ljharb: Since this calls into the original function, shouldn't it already return `NaN`?

In which browsers/engines does the existing shim not do this?

ljharb: Also, `parseInt('undefined')` should also be `NaN`, so since `parseInt` changes all input to strings, checking the `typeof` isn't very helpful.

ValeryVS: In firefox 17 parseInt(null) and parseInt(undefined) are throw "can't convert undefined to object".
But parseInt(NaN) is ok. Need some testing.

Probably that is because of 
`var trim = call.bind(String.prototype.trim);`

ljharb: aha, ok - in that case, I probably just need `trim(String(str))`

ValeryVS: Yes, this will work.
Also, this problem is for undefined and null only.

`Function.prototype.call.bind(String.prototype.trim)(undefined)`
throws Uncaught TypeError: String.prototype.trim called on null or undefined

ValeryVS: Trim will convert any other variable to String, anyway. So, `trim(String(str))` seems to be nice solution.

ljharb:
If you want to change your implementation diff to be just what we discussed, and then rebase everything down to a commit on latest master, I'd love to merge it :-)

ValeryVS:
Done.

ljharb:
@ValeryVS hmm, I'm checking in FF 17, and even on master, all of your added tests are passing. Which version of es5-shim are you using? Can you compare the latest?

Can you create a jsfiddle or equivalent that reproduces the failure?

ValeryVS:
Here is plunk
https://plnkr.co/edit/OjdaQ9xbq66yg4zmxlFR

es5-shim file from master, but "if" conditions near parseInt and parseFloat are commented out.
So, even in modern browsers, there will be
"TypeError: String.prototype.trim called on null or undefined"
in console.

zloirock:
@ljharb I wrote you about this bug ~ half a year ago. One new report - https://github.com/zloirock/core-js/issues/208#issuecomment-224950441

ValeryVS:
FF 17 throws
TypeError: can't convert undefined to object
https://run.plnkr.co/4XDGZUobizVlg7AI/es5-shim.js
Row: 1983

ljharb:
@zloirock i don't see an issue filed - random comments on random threads invariably get lost.

@ValeryVS I get a 404 on that link.

ValeryVS:
@ljharb 
You, probably open link to script from error.

This link to plunk from my first comment works

> Here is plunk
> https://plnkr.co/edit/OjdaQ9xbq66yg4zmxlFR

To see the error, you must open browser console and ther click the "run" button.

ljharb:
@ValeryVS aha, got it :-) thank you!

PR #404, opened 2016-06-16
Add Array.prototype.find and Array.prototype.findIndex

whawker:
Possibly out of scope of this project, as they are ES6 array methods. But they are incredibly useful all the same.

[MDN Array.prototype.find docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)
[MDN Array.prototype.findIndex docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex)

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -568,6 +568,61 @@
         }
     }, !properlyBoxesContext(ArrayPrototype.some));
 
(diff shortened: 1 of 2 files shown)


ljharb:
These functions are already available in the [es6-shim](https://github.com/paulmillr/es6-shim), and definitely ES6 functions are out of scope of this project.

Issue #412, opened 2016-10-13
String.prototype.replace works incorrectly with RegExp assertion ?=


YLeight:
In case when we have regular expression with assertion `?=`, the `wrappedReplaceValue`, which decorate the original `replaceValue` function, can't define correct arguments, because the coincidence which was passed to `replaceValue` function don't match the RegExp pattern and the final arguments have just only two values - the offset of the matched sub-string within the whole string being examined, and the whole string being examined.
There is a certain problem, when whole RegExp don't match with found coincidence, but in fact, there is a coincidence [StringPrototype.replace polyfill](https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L1892)

ljharb:
Can you provide specific code (like a specific regex that has a lookahead)?

YLeight:
`/(\n)?(^[ \t]*)([*+-]|\d+[.])[ \t]+((\[(x|X| )?])?[ \t]*[^\r]+?(\n{1,2}))(?=\n*(~0|\2([*+-]|\d+[.])[ \t]+))/gm`

YLeight:
[StringPrototype.replace polyfill](https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L1892)
If the `searchValue` equal `/(\n)?(^[ \t]*)([*+-]|\d+[.])[ \t]+((\[(x|X| )?])?[ \t]*[^\r]+?(\n{1,2}))(?=\n*(~0|\2([*+-]|\d+[.])[ \t]+))/gm` and `match` equal '* some list item;' we get null in 1892 code line but not the groups.

ljharb:
```js
var str = '* some list item;';
var re = /(\n)?(^[ \t]*)([*+-]|\d+[.])[ \t]+((\[(x|X| )?])?[ \t]*[^\r]+?(\n{1,2}))(?=\n*(~0|\2([*+-]|\d+[.])[ \t]+))/gm;
[str.match(re), re.exec(str), str.replace(re, function () { console.log(arguments); })]
```
in the native browser console implies there should be no match. Can you provide a specific test that passes on a compliant browser but fails on a browser where the es5-shim has polyfilled `String#replace`?

YLeight:
I mean next case:
``` javascript
[...]

PR #413, opened 2016-10-21
es5-shim.js: fix issue #412

YLeight:
Pass arguments to the original function

I have noticed that this [decorator](https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L1888) does not correct [problems](https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L1873) that are checked.

Background and reasons are:
In [1878 line](https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L1878) it is checked, that in the case of a capturing group with quantifiers preventing its exercise, the matched text for a capturing group has to be `undefined`, but there is no any correction.
However, the only one thing that is needed here is to correct types of missing groups, and pass all arguments to the original function.

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -1885,13 +1885,13 @@
             if (!isFn || !hasCapturingGroups) {
                 return str_replace.call(this, searchValue, replaceValue);
             } else {
-                var wrappedReplaceValue = function (match) {
-                    var length = arguments.length;
-                    var originalLastIndex = searchValue.lastIndex;
-                    searchValue.lastIndex = 0;
-                    var args = searchValue.exec(match) || [];
-                    searchValue.lastIndex = originalLastIndex;
-                    pushCall(args, arguments[length - 2], arguments[length - 1]);
+                var wrappedReplaceValue = function () {
+                    var args = [];
+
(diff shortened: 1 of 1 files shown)


ljharb:
@YLeight which browsers have you tested this in?

YLeight:
I have tested:

- [ff 15](https://ftp.mozilla.org/pub/firefox/releases/15.0/linux-x86_64/en-US/)

- [ff 20](https://ftp.mozilla.org/pub/firefox/releases/20.0/linux-x86_64/en-US/)

- [ff 24](https://ftp.mozilla.org/pub/firefox/releases/24.0/linux-x86_64/ru/)

- [ff 33](https://ftp.mozilla.org/pub/firefox/releases/33.0/linux-x86_64/en-US/)

ljharb:
@YLeight are you no longer interested in getting this in?

YLeight:
This is not a great solution, because just groups with quantifiers preventing its exercise have to be undefined. Groups without quantifiers have to be empty string by standard. There are need more complex solution or somthing else. I try to solve this issue, and found one decision. When I will finish, I will create new pull request. And you can make a review.

ljharb:
Thanks!

PR #418, opened 2016-11-24
Issue#412

YLeight:
@ljharb I have added two tests which show thin places of both decisions in FF versions > 34 (current one and my previous one) [from line 39 on](https://github.com/YLeight/es5-shim/blob/issue%23412/tests/spec/s-string.js#L39)
My current solution is a littlle bit difficult, but I have not found more simple one yet.

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -1925,20 +1925,93 @@
     }());
 
     if (!replaceReportsGroupsCorrectly) {
+        /**
+         * @param regexp {RegExp}
+         * @returns {Object[]} groups
+         *  @return {number} groups[].number        - The number of group.
+         *  @return {boolean} groups[].mayBeMissing - The group with quantifiers preventing its exercise, the matched text
+         *                                            for a capturing group is now undefined instead of an empty string.
+         *                                            The group without quantifiers, may be empty string.
+         *  @return {boolean} groups[].isResulting  - The groups with ?:, ?=, ?! are non-capturing groups.
+         */
+        var getExtendedGroups = function getExtendedGroups(regexp) {
+            var source = regexp.source;
+            var result = [];
+            var groupNumberCounter = 0;
+
+            for (var charIndex in source) {
+                var char = source[+charIndex];
+                var prevFirst = source[+charIndex - 1] || null;
+                var nextFirst = source[+charIndex + 1] || null;
+                var nextSecond = source[+charIndex + 2] || null;
+
+                if (char === '(') {
+                    if (prevFirst && prevFirst === '\\') {
+                        continue;
+                    }
+
+                    if (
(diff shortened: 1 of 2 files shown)


es5-shim.js
```diff
@@ -1921,20 +1921,93 @@
...
+
+                if (char === ')') {
+                    var groupIndex = groupNumberCounter - 1;
+
+                    if (prevFirst && prevFirst === '\\') {
+                        continue;
+                    }
+
+                    if (result[groupIndex] && !result[groupIndex].isResulting) {
+                        result.splice(groupIndex, 1);
```
ljharb: we should definitely not rely on:
 - `splice` working properly
 - `splice` being present

nor should we mutate the array like this.

es5-shim.js
```diff
@@ -1921,20 +1921,93 @@
...
             } else {
                 var wrappedReplaceValue = function (match) {
-                    var length = arguments.length;
-                    var originalLastIndex = searchValue.lastIndex;
-                    searchValue.lastIndex = 0; // eslint-disable-line no-param-reassign
-                    var args = searchValue.exec(match) || [];
-                    searchValue.lastIndex = originalLastIndex; // eslint-disable-line no-param-reassign
-                    pushCall(args, arguments[length - 2], arguments[length - 1]);
-                    return replaceValue.apply(this, args);
+                    var groups = ArrayPrototype.slice.call(arguments, 1, -2);
```
ljharb: `delete Array.prototype.slice` would break this code.
```suggestion
                    var groups = arraySlice(arguments, 1, -2);
```

es5-shim.js
```diff
@@ -1921,20 +1921,93 @@
...
+
+                    for (var i = 0; i < groups.length; i++) {
+                        var argumentIndex = i + 1;
+                        if (groups[i] === '' && extendedGroups[i].mayBeMissing) {
+                            arguments[argumentIndex] = undefined;
```
ljharb: assigning to `arguments` is a massive deoptimization, and is mutation anyways. we'll have to avoid this.

es5-shim.js
```diff
@@ -1921,20 +1921,93 @@
...
+                        }
+                    }
+
+                    return replaceValue.apply(this, arguments);
```
ljharb:
```suggestion
                    return apply.call(replaceValue, this, arguments);
```

ljharb (changes requested):
@YLeight sorry for the long delay here.

I've rebased this and cleaned up a number of lint violations. However, there's still some left, plus some additional comments.

ljharb:
ping @YLeight; i've rebased this, but there's still linting issues as well as mutation.

PR #445, opened 2017-06-26
🐛 Fix: Accessing window.top.{constructor|prototype} from a frame throws error in iOS

tomgasson:
Prevent accessing `window.top` so that no security errors are thrown when es5-shim is used in an iframe from a different domain


tomgasson:
I added the test

PR #455, opened 2018-02-18
Check for module's existence instead of exports

MelkorNemesis:
The library should be checking for `module` existence instead of `exports` existence. It has happened to us that in the browser for whatever reason `exports` was pointing to `Window` and the module check failed on `Module is not defined` on line 25 in this file. Also it doesn't make sense to check for variable `A` and then writing to variable `B` (I know they're aliases in Node, but that's not the case in browser).

### DIFF
--- a/es5-shim.js
+++ b/es5-shim.js
@@ -18,7 +18,7 @@
     if (typeof define === 'function' && define.amd) {
         // AMD. Register as an anonymous module.
         define(factory);
-    } else if (typeof exports === 'object') {
+    } else if (typeof module === 'object') {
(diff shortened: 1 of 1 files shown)


borisvalo:
+1

ljharb:
This is a standard UMD that has worked, unchanged, for the better part of a decade. Are you using es5-shim directly, or as part of a bundler? If directly, what was setting the global exports variable? That’s likely to break a good many libraries you might be using.

Separately, This isn’t the right fix, because `module = Window` would break your code too, would it not?

